PR #2, opened 2023-02-04
cleanup

nordfjord:
- token and state
- naming
- async enumerables
- fix

### DIFF
--- a/packages/core/src/lib/Cache.ts
+++ b/packages/core/src/lib/Cache.ts
@@ -1,4 +1,4 @@
-import { StreamToken } from "./Core"
+import { StreamToken, TokenAndState } from "./Core"
 import LRUCache from "lru-cache"
 
 type Expiration = { absolute: number } | { relative: number }
@@ -12,12 +12,12 @@ export class CacheEntry<State> {
       this.state = other.state
     }
   }
-  value(): [StreamToken, State] {
-    return [this.token, this.state]
+  value(): TokenAndState<State> {
+    return { token: this.token, state: this.state }
   }
 
-  static ofTokenAndState<S>([token, state]: [StreamToken, S]) {
-    return new CacheEntry(token, state)
+  static ofTokenAndState<S>(x: TokenAndState<S>) {
+    return new CacheEntry(x.token, x.state)
   }
 }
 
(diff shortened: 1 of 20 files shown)


packages/core/src/lib/Cache.ts
```diff
@@ -12,12 +12,12 @@ export class CacheEntry<State> {
       this.state = other.state
     }
   }
-  value(): [StreamToken, State] {
-    return [this.token, this.state]
+  value(): TokenAndState<State> {
```
bartelink, line 15: 🤔 wonder would using a struct record help in the base impl - probably not as it will make some stuff more longwinded...

nordfjord: > wonder would using a struct record help in the base impl

One reason I'm avoiding tuples is because destructuring tuples incurs the creation of an iterator in most JS runtimes. I happen to find records nicer in general, but there's a performance cost in JS for tuples that doesn't exist in .NET

nordfjord: Some anecdata for this preference:

https://github.com/jet/equinox/blob/master/src/Equinox.DynamoStore/DynamoStore.fs#L615
https://github.com/jet/equinox/blob/master/src/Equinox.DynamoStore/DynamoStore.fs#L644

Notice the difference in names from the type and the variable passed?

https://github.com/jet/equinox/blob/master/src/Equinox.DynamoStore/DynamoStore.fs#L693-L700

Notice that in mkQuery we have `maxItems (direction, minIndex, maxIndex)` but in `mapPage` `(minIndex, maxIndex, maxItems)`

Records keep this stuff consistent so it's easier (for me) to navigate the code

bartelink: Thanks for the perf side point.

I accept the first point on the consistency of naming (that point is also relevant in the second one - maxItems is aliased as batchSize for debatable reasons - IIRC the esdb impl calls them batch size, Cosmos tries to use items to align with its SDK terms, DDB follows suit as Items is the relevant term there too, but the serilog field is different)

Not sure I agree with the `maxItems` ordering/tupling consistency point - in the context of mapPage, they are side-info for logging.
In the context of `mkQuery`, they are separate aspects of the query (the actual query params vs batching hints). There are also frequently differences in optionality etc

In this case a spike says they can be made consistent without any major mess (probably by grouping and ordering the tupled bits consistently), but you can definitely go too far with making named records for arbitrary things that happen to be common to say 5 call sites and 2 signatures. Ultimately IMO having a named record for everything decreases the value of the others and their names.

Would be happy to take PRs for any cleanups you identify, especially if they also work well in the context of the JS codebase (though attempting to keep them in step forever is obviously not a good use of time or focus)

PR #3, opened 2023-02-04
add CI pipeline

nordfjord:
cc @bartelink.

This file should be easily modifiable to run the dynamo integration tests for equinox 😉


bartelink:
Nice; only a matter of time now :D

PR #7, opened 2023-07-25
fix build

oomathias:
Consolidate build.

Fix `TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("@equinox-js/message-db-consumer")' call instead.`


nordfjord:
Thank you

PR #8, opened 2023-07-27
feat: support sql condition for category reading

nordfjord:
Support the breadth of message-db options


apps/docs/docs/message-db/source.md
```diff
@@ -0,0 +1,61 @@
+# Source
+
+The MessageDB source continuously fetches batches of messages and dispatches
```
bartelink, line 3: redundant "and dispatches them to a supplied handler."

apps/docs/docs/message-db/access-strategies.md
```diff
@@ -1,18 +1,23 @@
 # Access Strategies
 
+An access strategy controls how your category will _access_ events from the
```
bartelink, line 3: I hate undocumentation sentences

The access strategy allows you to opt into custom loading or reloading algorithms. For almost all categories in almost all most systems, `Unoptimized` is a good choice; you can always switch later.

apps/docs/docs/message-db/source.md
```diff
@@ -0,0 +1,61 @@
...
+   * @default 500
+   */
+  batchSize?: number
+  /** The name of the consumer group to use for checkpointing */
+  groupName: string
+  /** The checkpointer to use for checkpointing */
+  checkpointer: ICheckpointer
+  /** The handler to call for each batch of stream messages */
+  handler: (streamName: string, events: ITimelineEvent<string>[]) => Promise<void>
+  /** The number of milliseconds to sleep between tail reads */
```
bartelink, line 29: more ai at work!

sleep time in ms, when reading has reached the end of the category's events

apps/docs/docs/message-db/source.md
```diff
@@ -0,0 +1,61 @@
...
+1. The checkpointer will checkpoint to `{groupName}-{member}`
+2. MessageDB will filter messages based on a hash of the stream id
+
+This means that each consumer in the group maintains a distinct checkpoint and
+is therefore not easily autoscaled. In order to expand or contract a consumer
+group you will need to wipe the checkpoints and have all of the consumers in the
+group start from the current lowest checkpoint.
+
+Using a hash of the ID guarantees that all messages for a particular entity will
+be handled by the same consumer.
```
bartelink, line 61: be handled in the correct order by the same consumer instance

(I think using stream instead of entity here will help convey what they need to be thinking)

nordfjord: Message db uses the "cardinal id" to do its sharding, Cardinal ID being based on the eventide concept where a stream name is formed as `{category}-{cardinal_id}+{other_ids}` so we kind of opt out of the behaviour by using `_` as our stream id separator. But in the case where you have e.g. an `Invoice-1` and an `InvoiceAutoEmailer-1` you would be guaranteed that they're both handled by the same consumer instance and therefore would benefit from the caching layer most likely having an up-to-date view. Or at least that's how I imagine Scott's reasoning for the behaviour goes.

That said, maybe it's not worth calling out here

bartelink: Interesting. Can't think of a thing I've done or wanted to do which would benefit enough from this to warrant attempting to convey the concept in a terse form here. Ultimately if your dispatch is at stream level and concurrent, the concept is not relevant here. But you do want to convey that all messages will consistently route to the same consumer instance and hence guarantee no concurrent processing for a given stream.
Either way, overindexing on the term entity without layout out the concept in this context is not something to overindex on. While I'm sure it will speak to an audience, I still feel that entity is equally likely to get people thinking CRUD or "sending messages: rather than getting them in the right frame of mind for handling coalesced groups of events at positions in a sequential stream (I see lots of cases where people see the goal of an event contract/structure to be to double as a message that can be handled in one shot by a consumer. I'm all for using direct approaches that solve problems over tonnes of layers that some dogmatic best practice scheme suggests, but you still need to default to talking in terms that are helpful to conveying a good general strategy).

nordfjord: agreed, I'll remove the reference to entity, I think it'll cause more harm than good

PR #10, opened 2023-07-30
Optimised SQL read models

nordfjord:
# Use case

We sometimes want to create SQL read models and so far I've been asking people to write out SQL manually. This is a fine way to get started but can lead to a lot of code.

Here we introduce a projection library that allows you to specify updates as `Insert`, `Update`, `Delete`, and `Upsert`. The library takes care of intelligently collapsing change sets utilising the guarantee that each call to the handler is for a single stream and therefore a single entity and therefore is safely collapsible. This library is not meant for use cases where that does not hold.

This optimises the number of roundtrips required to the database server to 1 for each changeset.

## Example use

```ts
function changes(streamName: string, events: ITimelineEvent<string>[]): Change[] {
   const id = AppointmentId.parse(StreamName.parseId(streamName))
   const decodedEvents = keepMap(events, Appointment.codec.tryDecode)
   const result: AppointmentChange[] = []
   for (const event of decodedEvents) {
     switch (event.type) {
       case "AppointmentScheduled":
         result.push(
           Insert({ id, title: event.data.title, start: event.data.start, duration_ms: event.data.duration_ms, is_cancelled: false }))
       case "AppointmentRescheduled":
         result.push(Update({ id, start: event.data.start }))
       case "AppointmentCancelled":
         result.push(Update({ id, is_cancelled: true }))
     }
   }
   return result
 }
```

# Todo

- [ ] Support idempotency via a version column

### DIFF
--- a/packages/projection-pg/src/collapse.ts
+++ b/packages/projection-pg/src/collapse.ts
@@ -1,5 +1,7 @@
 import { Action, Change, Insert, Update, Upsert } from "./types"
 
+// Note: do not run prettier over this file, it does a bad job and makes it harder to read
+
 export function collapseChanges(changes: Change[]): Change | undefined {
   if (changes.length <= 1) return changes[0]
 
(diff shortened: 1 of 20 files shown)


packages/projection-pg/src/collapse.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export function collapseChanges(changes: Change[]): Change | undefined {
+  if (changes.length <= 1) return changes[0]
+
+  let result = changes[0]
+
+  for (let i = 1; i < changes.length; i++) {
+    const res = concat(result, changes[i])
+    // an undefined means we've zeroed out the change so we skip
+    if (res == null) { result = changes[i + 1]; i++; continue; }
```
bartelink, line 46: oh come on, this was an opportunity to enjoy a postincrement!

PR #11, opened 2023-07-31
Support MaxStaleMs load option

nordfjord:
- version bump
- feat: support max stale ms for cached reads

### DIFF
--- a/apps/store-integration/src/message-db.test.ts
+++ b/apps/store-integration/src/message-db.test.ts
@@ -24,7 +24,7 @@ namespace CartService {
   const fold = Cart.Fold.fold
   const initial = Cart.Fold.initial
   const cache = new MemoryCache()
-  const noCache = CachingStrategy.noCache()
+  const noCache = CachingStrategy.NoCache()
 
   export function createWithoutOptimization(context: MessageDbContext) {
     const category = Category.create(
@@ -48,7 +48,7 @@ namespace CartService {
     return Cart.Service.create(category)
   }
 
-  const sliding20m = CachingStrategy.slidingWindow(cache, 20 * 60 * 1000)
+  const sliding20m = CachingStrategy.Cache(cache)
 
   export function createWithCaching(context: MessageDbContext) {
     const category = Category.create(context, Cart.Category, codec, fold, initial, sliding20m)
--- a/docs/docs/core-concepts/caching.md
+++ b/docs/docs/core-concepts/caching.md
@@ -1,47 +1,70 @@
 # Caching
 
-Out of the box EquinoxJS supports a few strategies for caching. Caching in this context means that we'll keep 
-a record of aggregate state and it's version in memory. The `MemoryCache` included in the core library is a simple
-LRU cache which can keep up to a million states (configurable).
+Out of the box EquinoxJS supports an in-memory LRU cache. This cache is based
+on the lru-cache library. The default configuration allows 1,000,000 items to
+live in the cache. Since it's impossible to know the size of a JS object in
+bytes at runtime we cannot provide a sensible default for the cache that limits
+the actual memory usage of your process. You'll need to fine tune it to your
+actual requirements. As an example if you have 1 million states, each of which
+is 50 bytes that's 50MB of memory.
 
-On top of the cache we have a few Caching strategies to choose from
+The cache acts as a concurrency limiter on stream loads and guarantees that
+for each stream at most one load is happening concurrently.
 
-# NoCaching
+For performance reasons we do not offer TTL functionality on the cache. Doing
+so with the guarantees we'd want is not feasible with the underlying cache
+library. In particular, the underlying library will not pre-emptively evict
+expired items from the cache and as such your items would remain in the cache
+until requested again, at which point they'd be evicted. As such caching in
+EquinoxJS is a binary on/off setting.
+
+## `NoCaching`
 
 ```ts
 import { CachingStrategy } from "@equinox-js/core"
-const strategy = CachingStrategy.noCache()
+const strategy = CachingStrategy.NoCache()
 ```
 
 No caching is applied. This necessitates a fresh load of the entire stream for every interaction with the
 store.
 
-# SlidingWindow(cache, ms, ?prefix)
+## `Cache(cache, ?prefix)`
 
 ```ts
 import { CachingStrategy, MemoryCache } from "@equinox-js/core"
 const cache = new MemoryCache()
-const strategy = CachingStrategy.slidingWindow(cache, 20 * 60 * 1000)
+const strategy = CachingStrategy.Cache(cache)
 ```
 
-Retain a single `state` per `stream_name`.
-Each cache hit for a stream renews the retention period for the defined `ms`.
-Upon expiration of the defined `ms` from the point at which the cache was entry was last used, a full reload is
-triggered.
-Unless `LoadOption.AllowStale` is used, each cache hit still incurs a roundtrip to load any subsequently-added events.
+Retain a single `state` per `stream_name` in the supplied cache.
 
-Prefix is used to segregate multiple folds per stream when they are stored in the cache. 
+Prefix is used to segregate multiple folds per stream when they are stored in the cache.
 
-# FixedTimeSpan(cache, periodMs)
+# LoadOptions
 
-```ts
-import { CachingStrategy, MemoryCache } from "@equinox-js/core"
-const cache = new MemoryCache()
-const strategy = CachingStrategy.fixedTimeSpan(cache, 20 * 60 * 1000)
-```
+When transacting or querying on your decider you can supply a load option to
+opt into different load guarantees.
 
-Retain a single `state` per `stream_name`.
-Upon expiration of the defined `periodMs`, a full reload is triggered.
-Unless `LoadOption.AllowStale` is used, each cache hit still incurs a roundtrip to load any subsequently-added events.
+## `RequireLoad`
 
-Prefix is used to segregate multiple folds per stream when they are stored in the cache. 
+This is the default load option, always obtaining the latest state from the store
+
+## `RequireLader`
+
+Request that the stream be loaded with a quorum read / from a leader connection
+
+## MaxStale(ms)
+
+If the cache contains a state from the stream that was stored less than `ms`
+milliseconds ago that state will be used without fetching more events from the
+stream.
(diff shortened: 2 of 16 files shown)


apps/store-integration/src/message-db.test.ts
```diff
@@ -48,7 +48,7 @@ namespace CartService {
     return Cart.Service.create(category)
   }
 
-  const sliding20m = CachingStrategy.slidingWindow(cache, 20 * 60 * 1000)
+  const sliding20m = CachingStrategy.Cache(cache)
```
bartelink, line 51: fix name!

docs/docs/core-concepts/caching.md
```diff
@@ -1,47 +1,70 @@
 # Caching
 
-Out of the box EquinoxJS supports a few strategies for caching. Caching in this context means that we'll keep 
-a record of aggregate state and it's version in memory. The `MemoryCache` included in the core library is a simple
-LRU cache which can keep up to a million states (configurable).
+Out of the box EquinoxJS supports an in-memory LRU cache. This cache is based
+on the lru-cache library. The default configuration allows 1,000,000 items to
```
bartelink, line 4: [`lru-cache`](https://www.npmjs.com/package/lru-cache)

docs/docs/core-concepts/caching.md
```diff
@@ -1,47 +1,70 @@
...
+live in the cache. Since it's impossible to know the size of a JS object in
+bytes at runtime we cannot provide a sensible default for the cache that limits
+the actual memory usage of your process. You'll need to fine tune it to your
+actual requirements. As an example if you have 1 million states, each of which
+is 50 bytes that's 50MB of memory.
 
-On top of the cache we have a few Caching strategies to choose from
+The cache acts as a concurrency limiter on stream loads and guarantees that
```
bartelink, line 11: maybe "the cache doubles as"...

docs/docs/core-concepts/caching.md
```diff
@@ -1,47 +1,70 @@
...
+
+## `NoCaching`
 
 ```ts
 import { CachingStrategy } from "@equinox-js/core"
-const strategy = CachingStrategy.noCache()
+const strategy = CachingStrategy.NoCache()
 ```
 
 No caching is applied. This necessitates a fresh load of the entire stream for every interaction with the
```
bartelink, line 28: necessitates -> triggers

docs/docs/core-concepts/caching.md
```diff
@@ -1,47 +1,70 @@
...
-const strategy = CachingStrategy.fixedTimeSpan(cache, 20 * 60 * 1000)
-```
+When transacting or querying on your decider you can supply a load option to
+opt into different load guarantees.
+
+## `RequireLoad`
+
+This is the default load option, always obtaining the latest state from the store
+
+## `RequireLader`
```
bartelink, line 52: RequireLadder

docs/docs/core-concepts/caching.md
```diff
@@ -1,47 +1,70 @@
...
+
+Request that the stream be loaded with a quorum read / from a leader connection
+
+## MaxStale(ms)
+
+If the cache contains a state from the stream that was stored less than `ms`
```
bartelink, line 58: stored -> read or written

docs/docs/core-concepts/caching.md
```diff
@@ -1,47 +1,70 @@
...
+milliseconds ago that state will be used without fetching more events from the
+stream.
+
+## `AnyCachedValue`
+
+If the cache contains any state from the stream, that state will be used without
+fetching more events from the stream.
```
bartelink, line 65: Useful to avoid a read roundtrip if you believe there is a good chance that the value in the cache will be correct (if not, and append will trigger a reload due to a concurrency conflict during the <sync operation, whatever you term it in here>)

packages/core/src/lib/Cache.ts
```diff
@@ -4,158 +4,121 @@ import { ICategory } from "./Category.js"
...
+    skipReloadIfYoungerThanMs: number,
+    read: (tokenAndState?: TokenAndState<State>) => Promise<TokenAndState<State>>,
+  ): Promise<TokenAndState<State>>
 
-  tryGet<State>(key: string): Promise<TokenAndState<State> | undefined>
+  updateIfNewer<State>(key: string, entry: CacheEntry<State>): void
 }
 
-type Supersedes = (a: StreamToken, b: StreamToken) => boolean
+export const supersedes = (current: StreamToken, x: StreamToken) => x.version > current.version
```
bartelink, line 40: if an entry is re-read, you want to touch the entry if it is `=` (have not read full thing yet tho)

packages/core/src/lib/Cache.ts
```diff
@@ -4,158 +4,121 @@ import { ICategory } from "./Category.js"
...
+   * It's not possible to know the size of a JS object in bytes at runtime,
+   * as such we cannot provide a sensible default for the cache that limits
+   * the actual memory usage of your process. You'll need to fine tune it to your
+   * requirements. By default we allow 1 million entries in the cache.
+   * If each entry is 50 bytes that's 50MB of memory.
+   *
+   * @param max Maximum number of entries to cache
+   */
+  constructor(max = 1_000_000) {
+    this.cache = new LRUCache({ max, ttl: 0 })
```
bartelink, line 54: one liner about reason for choosing TTL 0

packages/core/src/lib/Cache.ts
```diff
@@ -4,158 +4,121 @@ import { ICategory } from "./Category.js"
...
+  }
+
+  fetchers: Map<string, Promise<any>> = new Map()
```
bartelink, line 57: loadsInProgress / wipo

packages/core/src/lib/Cache.ts
```diff
@@ -4,158 +4,121 @@ import { ICategory } from "./Category.js"
...
+
+  private readWithExactlyOneFetch<State>(key: string, read: () => Promise<TokenAndState<State>>) {
```
bartelink, line 59: coalesceConcurrentReads

packages/core/src/lib/Cache.ts
```diff
@@ -4,158 +4,121 @@ import { ICategory } from "./Category.js"
...
+    public state: T,
+    public cachedAt: number,
   ) {}
 
-  updateIfNewer(supersedes: (a: StreamToken, b: StreamToken) => boolean, other: CacheEntry<State>) {
+  updateIfNewer(other: CacheEntry<T>) {
     if (supersedes(this.token, other.token)) {
       this.token = other.token
       this.state = other.state
+      this.cachedAt = Date.now()
```
bartelink, line 18: this needs to move out of the condition, i.e. a re-read needs to freshen the age

packages/core/src/lib/Cache.ts
```diff
@@ -4,158 +4,121 @@ import { ICategory } from "./Category.js"
...
     }
   }
-  value(): TokenAndState<State> {
+  value(): TokenAndState<T> {
     return { token: this.token, state: this.state }
   }
 
   static ofTokenAndState<S>(x: TokenAndState<S>) {
-    return new CacheEntry(x.token, x.state)
+    return new CacheEntry(x.token, x.state, Date.now())
```
bartelink, line 26: is this def the most efficient way to compute an instant? the actual date is not of relevance and e.g. we are not interested in daylight savings or clock smoothing etc (actually thinking of how heavy a date API can be triggerd by a recollection of it dominating a profile in C++!)

packages/core/src/lib/Cache.ts
```diff
@@ -4,158 +4,121 @@ import { ICategory } from "./Category.js"
...
-    supersedes: Supersedes,
-    entry: CacheEntry<State>,
-  ): Promise<void> {
-    const ttl = "absolute" in expiration ? expiration.absolute - Date.now() : expiration.relative
+  updateIfNewer<State>(key: string, entry: CacheEntry<State>): void {
     if (!this.cache.has(key)) {
-      this.cache.set(key, entry, { ttl })
+      this.cache.set(key, entry)
     } else {
       const current = this.cache.get(key)!
```
bartelink, line 101: would it really kill them to offer a tryget?

packages/core/src/lib/Cache.ts
```diff
@@ -4,158 +4,121 @@ import { ICategory } from "./Category.js"
...
-  ): ICachingStrategy => new SlidingWindow(cache, windowInMs, prefix)
-  export const noCache = (): ICachingStrategy => new NoCaching()
-  export const fixedTimeSpan = (
-    cache: ICache,
-    periodInMs: number,
-    prefix?: string,
-  ): ICachingStrategy => new FixedTimeSpan(cache, periodInMs, prefix)
+  export const Cache = (cache: ICache, prefix?: string): ICachingStrategy => ({
+    cache,
+    cacheKey: (streamName: string) => (prefix || "") + streamName,
```
bartelink, line 119: `prefix === null ? streamName : <concat>` unless you are completly sure this is mapped more efficiently in all cases, given that no prefix is the 99.99% case

packages/core/src/lib/Cache.ts
```diff
@@ -166,31 +129,19 @@ export class CachingCategory<Event, State, Context> implements ICategory<Event,
   ) {}
 
   private cacheKey(streamId: string) {
-    return `${this.categoryName}/${streamId}`
+    return this.strategy.cacheKey(`${this.categoryName}/${streamId}`)
```
bartelink, line 132: if you can borrow a pre-concatenated form from the read path, it will reduce garbage

apps/store-integration/src/message-db.test.ts
nordfjord: `CachingStrategy.JustDoIt(cache)`?

bartelink: more the sliding20m aspect

packages/core/src/lib/Cache.ts
nordfjord: `Date.now()` returns a unix timestamp in ms. I'll run some micro benchmarks but pretty sure it's fine™️

apps/store-integration/src/message-db.test.ts
nordfjord: 🤦

PR #14, opened 2023-08-02
spike: include helper for stream name decoding

nordfjord:


packages/core/src/lib/StreamName.ts
```diff
@@ -12,6 +12,11 @@ export function parseCategory(streamName: string) {
   return streamName.slice(0, idx)
 }
 
+export function parseCategoryAndIds(streamName: string) {
```
bartelink, line 15: parseCategoryAndId at L5 should have a diff name to signify it does not parse and only splits
Also if there is no cat separator, does/should it throw or return undefined?

L1 compose is also problematic
- ideally category should be checked for embedded `-`
- and create is a better name as it just really glues them together

compose should string N elements together and add `_` seps (checking for embedded `_` in values)
- maybe it can be called join and parseIds can become split
- or is this stuff all called gen and the opposite is called parse?

Having a different name sets different expectations

Maybe doing a 3 liner intro at top of file about gen vs parse and create vs split or something might guide the naming?

Then, when parsing, there should be:
- Inverse of create (something like split?) to yield a pair
- the thing that gets the category (parseCategory) could just be called category as it does not parse
- the thing that gets the streamId (parseId) could be called streamId as it does not parse
- inverse of compose is called parseIds, maybe it should be called split (esp if you add a join)

apps/example/src/domain/invoice.ts
```diff
@@ -1,190 +1,207 @@
...
-// Fold
-export type InvoiceState = {
-  amount: number
-  payer_id: PayerId
-  payments: Set<string>
-  amount_paid: number
+export namespace Stream {
+  export const CATEGORY = "Invoice"
+  export const streamId = StreamId.gen(InvoiceId.toString)
+  export const parseId = StreamName.dec(CATEGORY, InvoiceId.parse)
```
bartelink, line 10: is there no 'try` idiom? 
is `parse` understood to be a process that can fail/reject?

nordfjord: yeah, parse is understood as having the possibility to throw. E.g. the `z.string().uuid().parse('1')` would throw a `ZodError`

apps/example/src/domain/invoice.ts
```diff
@@ -1,190 +1,207 @@
...
+      }
+    }
+
+  export const finalize = (state: State): Event[] => {
     switch (state.type) {
       case "Initial":
         throw new Error("Invoice not found")
       case "Finalized":
-        throw new Error("Invoice is finalized")
+        return []
```
bartelink, line 137: 🥳

apps/example/src/domain/invoice.ts
```diff
@@ -1,190 +1,207 @@
...
     const decider = this.resolve(id)
-    return decider.transact(finalize)
+    return decider.transact(Decide.finalize)
   }
 
   readInvoice(id: InvoiceId) {
     const decider = this.resolve(id)
-    return decider.query(summary)
+    return decider.query(Query.summary)
   }
```
bartelink, line 190: create and resolvecategory and random things like defaulting / loading config are `Factory` things?
I guess it makes more sense if you have a enough aggregates...

apps/example/test/domain/scenario.ts
```diff
@@ -6,10 +6,13 @@ type ScenarioState<Event, State> = {
   error?: unknown
 }
 
-export function createBDD<Event, State>(
-  fold: (state: State, events: Event[]) => State,
-  initial: State,
-) {
+export function createBDD<Event, State>({
```
bartelink, line 9: establish ? (anything but 3 caps and talking about BDD)

apps/example/src/domain/invoice.ts
bartelink: ok but this is not throwing for malformed - it's seeing whether the category matches ? in ruby there would be a `?` in the name ;)

nordfjord: > ok but this is not throwing for malformed

It is throwing on malformed, It's just not throwing if the category doesn't match

nordfjord: what do you call a function that can throw and returns an option? (don't say "stupid")

bartelink: tryParse
tryMatch
parseIf

apps/example/src/entrypoints/reactions.ts
```diff
@@ -22,7 +22,7 @@ const checkpointer = new PgCheckpoints(createPool(process.env.CP_CONN_STR)!)
 checkpointer.ensureTable().then(() => console.log("table created"))
 
 async function handle(streamName: string, events: ITimelineEvent<string>[]) {
-  const id = Invoice.Stream.parseId(streamName)
+  const id = Invoice.Stream.match(streamName)
```
bartelink, line 25: why is it not called tryDecode like the next call

PR #18, opened 2023-08-05
Mirror fscodec streamname changes

nordfjord:


apps/example/src/domain/payer.ts
```diff
@@ -5,9 +5,11 @@ import { equals } from "ramda"
 import * as Config from "../config/equinox.js"
 
 export namespace Stream {
-  export const CATEGORY = "Payer"
+  export const category = "Payer"
   export const streamId = StreamId.gen(PayerId.toString)
```
bartelink, line 9: `id` is not better, but it would be nice to find a name that nods to the streamid and decodeid being related
createId vs decodeId
genId vs decodeId
streamId vs decodeStreamId ?

PR #19, opened 2023-08-06
what if i made a schema lib

nordfjord:
Zod isn't really doing it for me as it's a one way validation library. For my use case I need a bidirectional transformation.


nordfjord:
yeah no, don't want to maintain this. Might copy it into a project if I ever need it

PR #21, opened 2023-08-21
fix intro.md

greg-cann:
Addresses an issue in the intro.md file to be consistent with the current implementation. 
- AccessStrategy is a function
- order of parameters do not match method signature
- forStream requires a context

### DIFF
--- a/docs/docs/intro.md
+++ b/docs/docs/intro.md
@@ -98,7 +98,7 @@ export function create(context: MessageDbContext, cache: ICache) {
     caching,
     access
   )
-  const resolve = (id: AccountId) => Decider.forStream(category, Stream.id(id), context)
+  const resolve = (id: AccountId) => Decider.forStream(category, Stream.id(id), null)
   return new Service(resolve)
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/docs/intro.md
```diff
@@ -95,10 +95,10 @@ export function create(context: MessageDbContext, cache: ICache) {
     Events.codec,
     Fold.fold,
     Fold.initial,
-    access,
     cache,
```
greg-cann, line 98: this should be `caching`
```suggestion
    caching,
```

docs/docs/intro.md
```diff
@@ -88,17 +88,17 @@ import { ICache, CachingStrategy, Decider } from "@equinox-js/core"
 
 export function create(context: MessageDbContext, cache: ICache) {
   const caching = CachingStrategy.Cache(cache)
-  const access = AccessStrategy.Unoptimized
+  const access = AccessStrategy.Unoptimized()
```
nordfjord, line 91: FYI: changed the call to here

PR #28, opened 2023-09-12
Dynamo store docs

nordfjord:


docs/docs/dynamo-store/access-strategies.md
```diff
@@ -0,0 +1,73 @@
+# Access Strategies
+
+The access strategy allows you to opt into custom loading or reloading
+algorithms. For almost all categories in almost all most systems, `Unoptimized`
```
bartelink, line 4: in almost all most

docs/docs/dynamo-store/access-strategies.md
```diff
@@ -0,0 +1,73 @@
...
+is a good choice; you can always switch later.
+
+
+# Unoptimized
+
+The first, and likely most used access strategy is the unoptimized one. It'll
+not apply any optimised reading logic. When loading a stream for the first time
```
bartelink, line 11: optimised -> optimized!

docs/docs/dynamo-store/access-strategies.md
```diff
@@ -0,0 +1,73 @@
...
+const toSnapshot = (state: State): Event => 
+  ({ type: snapshotEventType, data: { current: state } })
+```
+
+We can then create the access strategy
+```ts
+const access = AccessStrategy.Snapshot(isOrigin, toSnapshot)
+```
+
+With this in place Equinox will maintain and store a snapshot event in the tip
```
bartelink, line 70: maybe capital Snapshot

docs/docs/dynamo-store/access-strategies.md
```diff
@@ -0,0 +1,73 @@
...
+document that is guaranteed to always be up to date, and as a result the state
```
bartelink, line 71: item is prob better than document in DDB terms?
or maybe mention that there are conceptually pages/chapters, and you're looking in the conclusion chapter all the time

docs/docs/dynamo-store/access-strategies.md
```diff
@@ -0,0 +1,73 @@
...
+example could be a customer's contact preferences, or an entity view-data
+stream. This access strategy will ensure that you only load at most a single
+event when transacting or querying such a stream.
+
+# Snapshot(isOrigin, toSnapshot)
+
+Imagine a simple counter service. It does not have a defined start and end as
+most processes would and as such we don't know how large it'll grow over time.
+In order to ensure the snappiness of our application we might apply
+snapshotting to ensure that the state of the counter can be reconstructed in 2
```
bartelink, line 28: you dont circle back and mention that there's only one roundtrip though?

docs/docs/dynamo-store/access-strategies.md
```diff
@@ -0,0 +1,73 @@
...
+round-trips instead of N, where N is the number of events divided by batch
+size.
+
+In Equinox the snapshot is a member of the Event DU
+
+```ts
+type Event =
+  | { type: "Increment", data: { amount: number } }
+  | { type: "Snapshot", data: { current: number } }
```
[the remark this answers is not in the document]
bartelink: Noting it is used in one place below but should be used more

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
+---
+sidebar_position: 1
+---
+
+# DynamoStore Architecture
+
+When it comes to building an event store on top of a serverless NoSQL database
```
bartelink, line 7: lose the "it comes to"

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
...
+it is imperative to apply mechanical sympathy. We've seen implementations of
+DynamoDB event stores using a single-event-per-row strategy. This would seem the
```
bartelink, line 9: consider mentioning item/document alongside or instead of row?

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
...
+obvious choice so let's start by explaining why it's not the greatest idea.
+
+## DynamoDB cost model
+
+A great article to understand DynamoDB's pricing model was written by Zach
```
bartelink, line 14: Zach Charles has a great [article conveying the structure of DynamoDB's pricing model](https://zaccharles.medium.com/calculating-a-dynamodb-items-size-and-consumed-capacity-d1728942eb7c) - the charges closely mirror the actual work involved, and any good design will balance those forces well.

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
...
+Charles
+[here](https://zaccharles.medium.com/calculating-a-dynamodb-items-size-and-consumed-capacity-d1728942eb7c).
+
+The gist of it is that you pay per 4KB for reads, and 1KB for writes, and
+transactional writes will cost you double. As such DynamoStore is designed to
+utilise GetItem and PutItem APIs where possible and only does a table scan and
```
bartelink, line 20: backticks on the GetItem/PutItem
does a -> (drops down|resorts) to table scans and/or transactional batches where there's no other way.

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
...
+transactional write when absolutely necessary.
+
+With these constraints enumerated it should be clear why an event-per-row
+schema is not the greatest idea
+
+1. Reading a stream necessitates a scan in which you pay 0.5RUs at minimum for
```
bartelink, line 26: I agree with the point but I dont think you actually pay per row so people won't accept the point - it is charged based on the aggregated size. (In Cosmos I'd argue less, the headers are big and always travel but in DDB they dont)

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
...
+   each event in the stream
+2. Writing will always be transactional so you'll pay double
+
+## The architecture
+
+At heart DynamoDB is a document store and an event sourced system is composed
+of multiple streams. It therefore stands to reason that we should model the
+storage such that the stream is the document. This works great until we run into
```
bartelink, line 34: "the stream is the document" -> "that the events of the stream are co-located in a document"
I'd also consider adding a sentence that says that in a doc store you want to chop things into units that you want/need to write and/or read (and that for event sourcing, there is no case where you need to load/update/insert one event independent of all others).

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
...
+the 400KB max item size. To solve for this we will move the current events to
```
bartelink, line 35: move -> calve haha

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
...
+another document once the document size reaches a configurable threshold (in
+events or bytes). Streams are generally short-lived so calving should be a rare
+occurrence. For streams where it's unavoidable we provide functionality to store
+a snapshot alongside the main stream document ensuring that the state of the
```
bartelink, line 39: alongside -> within
at most 1 -> in a single roundtrip

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
...
+2. Writing will always be transactional so you'll pay double
+
+## The architecture
+
+At heart DynamoDB is a document store and an event sourced system is composed
+of multiple streams. It therefore stands to reason that we should model the
+storage such that the stream is the document. This works great until we run into
+the 400KB max item size. To solve for this we will move the current events to
+another document once the document size reaches a configurable threshold (in
+events or bytes). Streams are generally short-lived so calving should be a rare
```
bartelink, line 37: Streams -> "The majority of streams in a well-designed system should be naturally limited to a low number of events (and aggregate size)" ?

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
...
+occurrence. For streams where it's unavoidable we provide functionality to store
+a snapshot alongside the main stream document ensuring that the state of the
+stream can be reconstructed in at most 1 round trip.
+
+To summarise:
+
+- A stream is represented as one or more documents in the store (a Tip document
+  and 0 or more calved batches of events)
```
bartelink, line 45: maybe batches -> pages in this ? Damn, realizing I called the batches - why did I do that! (Answer is that before events in tip, each write made a batch, and commit is defintiely a bad term)

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
...
+- Since attribute keys count towards item-size we use single letter key names
+
+## Table schema
+
+| Key         | Type                                | Description                                                    |
+| ----------- | ----------------------------------- | -------------------------------------------------------------- |
+| `p`artition | String                              | The stream name (hash key)                                     |
+| `i`ndex     | Number                              | Int32.MAX for `Tip` 0-N for calved batches (range key)         |
+| `b`ytes     | ?Number                             | Only exists in the Tip, number of bytes in predecessor batches |
+| `etag`      | ?String                             | Etag of latest update (used for OCC)                           |
```
bartelink, line 56: [OCC](https://en.wikipedia.org/wiki/Optimistic_concurrency_control)

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
...
+| `e`vents    | List(Map([Event](#event-schema)))   | [Events](#event-schema)                                        |
+| `c`ases     | List(String)                        | Event types                                                    |
+| `u`nfolds   | List(Map([Unfold](#unfold-schema))) | [Unfolds](#unfold-schema)                                      |
+| `a`ppends   | Number                              | The number of events appended in the latest update             |
+| `n`umber    | Number                              | The version of the stream                                      |
```
bartelink, line 61: n (also) stands for `nextIndex ` - the `i`ndex that the next page starts at. In tip, it happens to coincide with the 'version', but I'd argue that this should be covered separately - the important concept is pages of events that are ordered/connected by their i/n indices, with a well known value for Tip's `i` so we can address it via a point read

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
...
+
+<h2 id="event-schema">Event schema</h2>
+
+| Key             | Type    | Description                                                  |
+| --------------- | ------- | ------------------------------------------------------------ |
+| `t`imestamp     | String  | ISO8601 DateTime representing the time the event was written |
```
bartelink, line 67: when it was encoded in the app - better not to oversimplify
Not sure if you saw but I made it DateTimeOffset.UtcNow recently
might eb worth it to convey that it potentially includes/stores the offset?

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
...
+| `x`actionId     | ?String | The correlation id of the event                              |
+| wh`y`           | ?String | The causation id (wh`y` did this event occur)                |
+
+<h2 id="unfold-schema">Unfold schema</h2>
+
+| Key             | Type    | Description                                                  |
+| --------------- | ------- | ------------------------------------------------------------ |
+| `i`ndex         | Number  | The index at which the unfold was generated                  |
+| `c`ase          | String  | The event type of the unfold                                 |
+| `t`imestamp     | String  | ISO8601 DateTime representing the time the event was written |
```
bartelink, line 81: the unfold was generated?

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
...
+| `D`ata encoding | ?Number | the encoding used for the `d`ata attribute                   |
+| `d`ata          | ?Binary | The event payload                                            |
```
bartelink, line 83: the unfold/snapshot's content?

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
...
+| `M`eta encoding | ?Number | the encoding used for the `m`eta attribute                   |
+| `m`eta          | ?Binary | The event metadata                                           |
```
bartelink, line 85: maybe mention not typically used
though I guess that would dilute the message of unfolds are just events with an associated base `i`

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
...
+
+# Reactions
+
+Reactions are an often necessary part of event-sourced architectures. DynamoDB
```
bartelink, line 89: I'd try to flip this to convey that they are almost the point once you've gotten past the toy 101 samples - there's always people to notify, indexes to be maintained, or lifecycles to things humans care about. They're not a terrible thing you have to do. - they're a thing you _get_ to to decouple things

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
...
+offers DynamoDB Streams, a Change Data Capture (CDC) for your table. This CDC
+feed offers two guarantees
+
+- Each stream record appears exactly once in the stream
```
bartelink, line 93: does 'stream record' relate to anything anyone will come here with some concept of?
maybe "every change (insert, update, delete) creates a single record in the DDB stream"

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
...
+- For each item that is modified in a DynamoDB table, the stream records appear
+  in the same sequence as the actual modifications to the item.
+- The stream contains the last 24 hours of updates
+
+You'll notice that the second guarantee notably does not guarantee ordering
+across items within the same hash key, only for updates to a singular (hash,
+range) pair. For this reason every event written to DynamoStore will be first
+written to the Tip document.
```
bartelink, line 101: will -> must.

Was trying to formulate a sentence but failed.... ~Maybe add "; if we skip 'feeding it in through the tip', the record of inserts of calf items can emanate from a DDB Streams feed after records relating to appends to the tip that preceded it in terms of when the writes actually happened"~

Also you have not made any reference to the key tenets of an event store's projections/reactions interface:-
1. all events for a stream absolutely must be delivered in order with no gaps (at least once delivery can always happen, but a feed that can ever supply you event 3 of a stream before it has presented you with event 2 is worse than useless)
2. there must be a 100% guarantee of all events getting delivered. The alternative is programming by coincidence, littering it with nonsensical defensive 'just in case' code, and spending hours checking logs to see if by any chance some event just might have been missed 'for any reason'.

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
...
+
+The third guarantee means DynamoDB streams cannot be used for general purpose
+crawling of all events in the store. One of the main selling points of
+event-sourced architectures is that you are able to create a new read-model as
+if it had always been there by processing all events in the store. To enable
+this we construct an index. This index lives in a secondary table and is kept up
+to date by a lambda function that's wired up against the DynamoDB stream. It
```
bartelink, line 108: Lambda

docs/docs/dynamo-store/architecture.md
```diff
@@ -0,0 +1,117 @@
...
+You'll notice that the second guarantee notably does not guarantee ordering
+across items within the same hash key, only for updates to a singular (hash,
+range) pair. For this reason every event written to DynamoStore will be first
+written to the Tip document.
+
+The third guarantee means DynamoDB streams cannot be used for general purpose
+crawling of all events in the store. One of the main selling points of
+event-sourced architectures is that you are able to create a new read-model as
+if it had always been there by processing all events in the store. To enable
+this we construct an index. This index lives in a secondary table and is kept up
```
bartelink, line 107: I know secondary alludes to secondary index etc, but its a thing of its own. 'sister', 'associated', 'separate related' ?

bartelink, line 107: and is kept up to date by -> continually extends to reference every event the store table from which the DDB Stream is being produced ?

bartelink: Maybe also make the overall thing explicit by tying it up by saying something like

"this index structure can then be deterministically traversed by current or future readers in perpetuity - i.e. there's no retention window, and no notion of 'replaying events' triggering unintended consequences"

(and then reference the guarantees of no missed events and no gaps in stream deliveries)

I'm also pondering if its TMI to say that the only thing other than DDB streams is a JSON dump of all data into a S3 tarball that takes single to double digit minutes to produce. (this might help people to conceptualize that an index can be generated/regenerated at will. If you then want to go completely to town you could say that a checkpoint is only valid for a single index - i.e. there is a single deterministic WORM thing being produced, but the exact batching/retries of the ingestion appends will vary the exact way in which things are grouped)

docs/docs/dynamo-store/access-strategies.md
```diff
@@ -0,0 +1,73 @@
...
+The access strategy allows you to opt into custom loading or reloading
+algorithms. For almost all categories in almost all most systems, `Unoptimized`
+is a good choice; you can always switch later.
+
+
+# Unoptimized
+
+The first, and likely most used access strategy is the unoptimized one. It'll
+not apply any optimised reading logic. When loading a stream for the first time
+all of its events will be loaded in batches of `batchSize` (default: `500`).
```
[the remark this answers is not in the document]
nordfjord: This is copy pasted from the message db docs and my tired ass didn’t read it 🙈

docs/docs/dynamo-store/access-strategies.md
nordfjord: This was copy pasted from the message db docs, and my tired ass didn’t even read it 🙈

docs/docs/dynamo-store/architecture.md
nordfjord: Yep, will fix the time of the time 😅.

ISO8601 should convey that offsets may be in play just fine, no?

docs/docs/dynamo-store/access-strategies.md
nordfjord: C'mon man! Expected you to represent the UK here 😂

docs/docs/dynamo-store/access-strategies.md
nordfjord: Good point, my terminology is all over the place. Document/Item/Tip/Stream/... Should pick one and go with it

Maybe I should read https://www.fernandomc.com/posts/dynamodb-key-terms-explained/

docs/docs/dynamo-store/access-strategies.md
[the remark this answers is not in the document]
nordfjord: I like the idea of the snapshot module, not sure it's necessary to show in this particular doc though 🤔. Might be too much

docs/docs/dynamo-store/architecture.md
nordfjord: Yeah I'm wrong on this, DDB stores items with the same partition key together and you pay for the data returned from the query

docs/docs/dynamo-store/architecture.md
nordfjord: Should I reword and use `Item` instead of document, and say DDB is a KV store at heart?

docs/docs/dynamo-store/architecture.md
nordfjord: Calving is esoteric terminology, will stick with move here and introduce calving as the term later :)

docs/docs/dynamo-store/access-strategies.md
bartelink: If you'd mentioned supporting the British Isles, perhaps :P its more the heading on L8 that prompts this!

docs/docs/dynamo-store/access-strategies.md
bartelink: should standardise on partition and sort per https://stackoverflow.com/questions/12241235/dynamodb-query-versus-getitem-for-single-item-retrieval-based-on-the-index#comment131304321_12248430
and if you're bored you can make sure the FS versions does that too!

docs/docs/dynamo-store/architecture.md
bartelink: I'd say its more than just KV, and the point is to leverage all that implies
But I think that is a good way to put it - "As with any KV store, point reads of a single item at a time are its sweet spot"
I dont think there'a any harm in invoking the term document once early on
then, KV store
and from then on switch to Partition of colocated Items sorted by the SortKey
And maybe work in the fact that DDB presents two addressable things
- Partition, which can be scanned or queried
- Item, which can be transacted against
Anything else: multiple writes, scattered reads, transactional updates is going through an extra layer, which means jitter, latency and/or charges.

I guess this also leads onto explaining that DDB streams is a scalable tree that represents all writes/deletes to Items over the last 24h. The only guarantee you get is that updates to any individual items will come out the other end in the order they were applied. 

When you think about it/put it that way, one wouldn't have any expectation of any relationship between the proceeds of updates to multiple items 
in a transactional batch update as they pop out the other end

docs/docs/dynamo-store/architecture.md
bartelink: I prefer mystical ;) but, agreed

docs/docs/dynamo-store/architecture.md
nordfjord: terminology is ripped straight from https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.html

nordfjord: I'll add some more info before this about the feed

bartelink: yeah key thing is we have an overload on 'stream' that the rest of the world doesnt
and stream record means something if you're a DDB streams dude, but nothing if you're coming off the street

PR #29, opened 2023-09-12
dynamo local

nordfjord:
- local dynamo
- wire the signal
- local dynamo
- stabilising

### DIFF
--- a/apps/example/src/entrypoints/reactions.ts
+++ b/apps/example/src/entrypoints/reactions.ts
@@ -58,7 +58,7 @@ function createSource() {
         query: QueryOptions.create({}),
       })
       const checkpoints = DynamoCheckpoints.create(
-        config.context,
+        context,
         config.cache,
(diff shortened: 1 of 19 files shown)


apps/example/src/entrypoints/reactions.ts
```diff
@@ -58,7 +58,7 @@ function createSource() {
         query: QueryOptions.create({}),
       })
       const checkpoints = DynamoCheckpoints.create(
-        config.context,
+        context,
         config.cache,
         300, // 5 minutes
```
bartelink, line 63: 5 minutes what

bartelink: also assign categories to a `handledCategories` to show that it needs to align
and pass groupName as an arg to createSource
for ddb, move mode toward front of arg list and handler to the end to match order in which things are relevant

PR #31, opened 2023-09-12
rolling state docs

nordfjord:


docs/docs/dynamo-store/access-strategies.md
```diff
@@ -57,11 +58,11 @@ Snapshot and how to transform the current state into a Snapshot
 
 ```ts
 const isOrigin = (ev: Event) => ev.type === "Snapshot"
```
bartelink, line 60:
```suggestion
const isOrigin = (ev: Event) => ev.type === snapshotEventType
```

docs/docs/dynamo-store/access-strategies.md
```diff
@@ -70,3 +71,63 @@ With this in place Equinox will maintain and store a Snapshot event in the tip
 document that is guaranteed to always be up to date, and as a result the state
 of the Stream can be reconstructed with a single point-read.
 
+## RollingState(toSnapshot)
+
+`RollingState` will throw away events, only storing the snapshot. This can be
```
bartelink, line 76:
```suggestion
`RollingState` builds state from events, just like the other modes. The key difference is that it discards the events, only storing the snapshot. This can be
```

docs/docs/dynamo-store/access-strategies.md
```diff
@@ -70,3 +71,63 @@ With this in place Equinox will maintain and store a Snapshot event in the tip
...
+useful for cases where we do not care about the individual events that led up to
+a state. This is frequently used to build up a read model that spans streams.
```
bartelink, line 78:
```suggestion
a state. Such a mechanism is useful for storing derived data built from things that you do have a log of events for, e.g. an summary that indexes data from across multiple streams. The main thing to bear in mind is that the total (compressed) size of the state is limited to the DynamoDB Item size limit of 400K.
```

docs/docs/dynamo-store/access-strategies.md
```diff
@@ -70,3 +71,63 @@ With this in place Equinox will maintain and store a Snapshot event in the tip
...
+    },
+  })
+}
+
+namespace Decide {
+  export const addUser = (id: string) => (state: State) => {
+    if (state.users.has(id)) return []
+    const newUsers = new Set(state.users)
+    newUsers.add(id)
+    return [{ version: state.version + 1, users: newUsers }]
```
bartelink, line 107: hm, does not manage deletion so not sure this is an ideal example 
if it did have to, then it would have to keep versions of each input stream

docs/docs/dynamo-store/access-strategies.md
bartelink: maybe mention that it does concurrency control as normal, and you're able to test it end to end without docker or a shared table?

docs/docs/dynamo-store/access-strategies.md
nordfjord: Why though? In TS  `ev.type` has type `"Snapshot" | "Increment"`

docs/docs/dynamo-store/access-strategies.md
nordfjord: What is this delete you speak of? 😜 

Do you have an idea for a better simple example?

docs/docs/dynamo-store/access-strategies.md
```diff
@@ -70,3 +71,63 @@ With this in place Equinox will maintain and store a Snapshot event in the tip
 document that is guaranteed to always be up to date, and as a result the state
 of the Stream can be reconstructed with a single point-read.
 
+## RollingState(toSnapshot)
+
+`RollingState` builds state from events, just like the other modes. The key difference is that it discards the events, only storing the snapshot. This can be
+useful for cases where we do not care about the individual events that led up to
+a state. Such a mechanism is useful for storing derived data built from things that you do have a log of events for, e.g. an summary that indexes data from across multiple streams. The main thing to bear in mind is that the total (compressed) size of the state is limited to the DynamoDB Item size limit of 400K.
```
bartelink, line 78:
```suggestion
a state. Such a mechanism is useful for storing derived data built from things that you already have a way to reason about, e.g. a summary that indexes data from across multiple streams. The main thing to bear in mind is that the total (compressed) size of the state is limited to the DynamoDB Item size limit of 400KiB.
```

docs/docs/dynamo-store/access-strategies.md
bartelink: good question
maybe track renames; that'll require keeping the ids and versions alongside the names

perhaps just keeping a map of ids to names and letting them cycle through all the superseded states in the context of handing a replay is not the end of the world

but I'd argue that you want to illustrate semi-interesting logic that you might want to test and/or not use native DDB update functions for

docs/docs/dynamo-store/access-strategies.md
bartelink: coz you used the var/constant in other places?
coz it should be `"Snapshotted"`?
to convey that it just gets renamed in all the places if you want to rev the format
hey wait you can't really do that, unless you upconvert from an older format, and typically you just change the category name for that case
hm,...

nordfjord: https://github.com/equinox-project/equinox-js/assets/3721521/765f88ba-4dae-4c8d-a41e-5dc9cbb065b1

This ain't you mama's type system 😉

docs/docs/dynamo-store/access-strategies.md
nordfjord: Updated to handle deletes

bartelink: ok, now the renames
... and  prevent it revisiting old states under replay!

nordfjord: Renames and all that landed

nordfjord: This all probably needs to be caveated as well. If you have 100K items in the map this'll just blow up. You could solve for it by linking to a `next` stream but even then, for a large enough dataset it should just become its own table.

(I'm guessing you're gonna school me about [epochs](https://github.com/jet/dotnet-templates/blob/master/equinox-patterns/README.md#list-epochsseries-with-exactly-once-ingestion-guarantee))

bartelink: Updated is prob a better event type name now
L113 maybe you still want to update the version?
maybe add a createMem
maybe prepend a `$` to cat name

bartelink: the caveat is implied in the 400K limit
no I can't think of a reasonable way to do it with epochs as the point of them is that they are append only and you walk them so they are fine to grow
for an index like this, you need to be able to reasonably
a) load into memory and do a manipulation in same
b) be able to serialize and transport the updated set
c) pay for an upateitem call without it getting daft

So if you have lots of items all getting updated the contention would get massive, and RU costs to update a 200K item to twiddle a few bytes get prohibitive (they're not that much better if you do a custom schema and use sets - its down to how big the item is at the end of the day, even if you can express the desired change as a tiny update command)

To deal with that contention, you might also use a Batcher if you had concurrency in play

you'd need to hash across a fixed number of docs to partition it, and then apply the update logic to a smaller subset

but then to load you'd need to either do a cross partition scan, or e.g. do N cached point reads (which is not as insane as it sounds if N is low)

the `eqx query` command I am due to add soon would provide an API to query across things
so e.g. on cosmos it would do a cross partitiion query for tips of things
and then can render a response direct from that
in the case of a deletable renameable item set that's within a reasonably limited set of items where the items have snapshots and/or all events are likely in tip, you'd then query the states and render as a list rather than go to the trouble of keeping an index
and you could also render that list as a paged result set with continuation tokens

Maybe you also should consider illustrating how you'd apply a consecutive set of upstream events batched into an array (either for one thing if you had a propulsion doing batches per upstream thing, or N events across M streams if you had that and a batcher to coordinate concurrent handings such as that). But that's probably another article

nordfjord: > the caveat is implied in the 400K limit

Glad we have tracing from which we can trigger alarms when a something like this grows to let's say 350K 😉 

> but then to load you'd need to either do a cross partition scan, or e.g. do N cached point reads (which is not as insane as it sounds if N is low)

I'd definitely do the N reads as a default.

> in the case of a deletable renameable item set that's within a reasonably limited set of items where the items have snapshots and/or all events are likely in tip, you'd then query the states and render as a list rather than go to the trouble of keeping an index

Where do you get the set of IDs to query?

nordfjord:
Merging this in for now, I think the example is good enough, and you're right, elaborating on strategies for larger sets is worthy of a separate post rather than a footnote under `RollingState`

docs/docs/dynamo-store/access-strategies.md
bartelink: >Where do you get the set of IDs to query?

set of ids to query is a prefix or pattern query based on the streamname - i.e. p startswith and i=-1
if you knew the ids, that's an async.parallelLimit or a BatchGetItems

docs/docs/dynamo-store/access-strategies.md
bartelink: replied and it got lost - there was a use of the constant further down - will do final check now...

docs/docs/dynamo-store/access-strategies.md
```diff
@@ -4,7 +4,6 @@ The access strategy allows you to opt into custom loading or reloading
 algorithms. For most systems, `Unoptimized` is a safe default; you can always
 switch later.
 
-
 # Unoptimized
 
 The first, and likely most used access strategy is the unoptimized one. It'll
```
bartelink, line 9: The first, and likely most used access strategy is the unoptimized one, which just scans to get the events across all the items in a Partition.

bartelink: >Loading uses a `Query` that loads batches backwards from the tip
Strange how this leaked into the docs; that only comes into play with isOrigin.

docs/docs/dynamo-store/access-strategies.md
bartelink: also "Snapshot" should be "Snapshotted" consistetly

PR #32, opened 2023-09-15
feat: Separate source and sink logic out of the sources

nordfjord:


packages/dynamo-store/source/src/DynamoStoreSource.mts
```diff
@@ -21,9 +20,9 @@ function keepMap<T, V>(arr: T[], fn: (x: T) => V | undefined): V[] {
 }
 
 type EventBody = EncodedBody
```
bartelink, line 22: still used?

packages/dynamo-store/source/src/DynamoStoreSource.mts
```diff
@@ -117,7 +112,7 @@ namespace Impl {
         // TOCONSIDER revise logic to share session key etc to rule this out
         const sliceFrom = span.i - Number(items[0].index)
         const events = items.slice(sliceFrom, sliceFrom + span.c.length)
-        for (const event of events) result.push([StreamName.parse(span.p), event])
+        for (const event of events) result.push([StreamName.parse(span.p), inflate(event)])
```
bartelink, line 115: hm if the inflate can be deferred, it might save memory consumption, memory churn and CPU if processing can filter based on event type or index

packages/dynamo-store/source/src/DynamoStoreSource.mts
```diff
@@ -274,65 +269,35 @@ function inflate(event: ITimelineEvent<EventBody>): ITimelineEvent {
...
-          await this.options.checkpoints.commit(
-            this.options.groupName,
-            AppendsPartitionId.toString(partition),
-            batch.checkpoint,
-          )
-          pos = batch.checkpoint
-        }
-        if (batch.isTail) await sleep(this.options.tailSleepIntervalMs, signal).catch(() => {})
-      }
+    if (!this.options.categories && !this.options.categoryFilter) {
```
bartelink, line 278: in cosmos I recently added a stream level filter - if there is no other usage of the category name in the lower level, better to map categories/categoryFilter/steamFilter to a stream predicate lambda at this level (though default surface API of a list of categories as a prominent default continues to be the main line)

packages/dynamo-store/source/src/DynamoStoreSource.mts
```diff
@@ -274,65 +269,35 @@ function inflate(event: ITimelineEvent<EventBody>): ITimelineEvent {
...
+      throw new Error("Either categories or categoryFilter must be specified")
     }
+    const sink = new StreamsSink(this.options.handler, this.options.maxConcurrentStreams)
+    const categoryFilter = options.categoryFilter ?? ((c) => options.categories!.includes(c))
```
bartelink, line 282: in cosmos I recently added a stream level filter - if there is no other usage of the category name in the lower level, better to map categories/categoryFilter/steamFilter to a stream predicate lambda at this level (though default surface API of a list of categories as a prominent default continues to be the main line)

packages/dynamo-store/source/src/DynamoStoreSource.mts
nordfjord: Actually, WTH, I've done it, see implementation of `inflate`

packages/dynamo-store/source/src/DynamoStoreSource.mts
nordfjord: could be more efficient to predicate that way 🤔 

```ts
streamName.startsWith(categoryName) && streamName[categoryName.length] === '-'
```

bartelink: interesting, I guess for the average category list, doing a few of these even beats splitting(and alocating) and then doing a set membership check

nordfjord: Screenshot 2023-09-16 at 14 55 07

So as expected, parsing the category is expensive, and a set with 2elems is slower than a linear search. Using a streamName filter with a for loop doing `startsWith(category + "-")` outperforms both by a factor of 2

nordfjord: Moved to stream filter

bartelink: nice. Should do same in F# (wonder whether your initial algorithm with the startswith and checking the char after the prefix is `-` might be more efficient there)...

PR #33, opened 2023-09-19
Support merging in future batches

nordfjord:


packages/propeller/src/Queue.ts
```diff
@@ -0,0 +1,37 @@
+class Node<T> {
```
nordfjord, line 1: Lol at JS not having a built in queue implementation 🥴

bartelink: This is why we get people to code important complex algorithms stuff in interviews

packages/propeller/src/StreamsSink.mts
```diff
@@ -16,38 +16,117 @@ function groupByStream(items: [StreamName, ITimelineEvent][]): Map<StreamName, I
...
+    if (!last || last.index === event.index - 1n) {
+      this.events.push(event)
+      return
+    }
+    if (last.index > event.index) return
+    if (event.index > last.index) throw new Error("Unexpected gap")
+  }
+}
+
+const getOrAdd = <K, V>(map: Map<K, V>, key: K, val: () => V) => {
```
bartelink, line 40: getelseadd would be the name I'd expect given val is a factory?

packages/propeller/src/StreamsSink.mts
```diff
@@ -16,38 +16,117 @@ function groupByStream(items: [StreamName, ITimelineEvent][]): Map<StreamName, I
...
+
+      for (let i = 0; i < this.maxConcurrentStreams; ++i) aux()
+    })
+  }
+
+  pump(batch: IngesterBatch, signal: AbortSignal) {
+    return new Promise<void>((resolve, reject) => {
+      this.inProgressBatches++
+      const names = new Set<StreamName>()
+      this.batchStreams.set(batch, names)
```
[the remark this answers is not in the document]
nordfjord: The thought process here is:

The queue is processed in order, and streams are added to the queue in batch-order.
We mutate the value of the stream in the queue when more batches come in with data for the same stream.
When we start processing a stream we remove it from the map, meaning that if another batch comes in for the same stream it'll add another `Stream` to the queue and map.

I think you're right that there's an edge case, but I think it's in the opposite direction. That is, it's possible to commit too early because we can remove the stream from the set before having actually handled all the events from that batch.

I think the solution is to use a set of `Stream` instead of `StreamName` 🤔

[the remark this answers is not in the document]
nordfjord: Let's get on a call soon so I can take you through the logic 😅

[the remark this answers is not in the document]
nordfjord: it boils down to on line 88 we go through each batch and handle the fact that the stream is processed. This means removing it from the set, and if the set becomes empty as a result completing the batch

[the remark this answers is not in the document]
nordfjord: I'll have to check the spec to see if the iteration order of a map is stable and in insertion order, if not I'll move to a ResizeArray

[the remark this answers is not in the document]
bartelink: Seems correct in general
Things are def simpler if handlers don't get to say "used half of them" and/or I failed; chuck them back in the pool for retry (merged with any that arived in the interim)
call won't fit today unfortunately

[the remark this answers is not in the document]
nordfjord: Was definitely not intending “soon” to mean today 😅

packages/message-db/message-db-consumer/src/lib/MessageDbSource.mts
```diff
@@ -21,8 +21,12 @@ interface CreateOptions {
   handler: (streamName: StreamName, events: ITimelineEvent[]) => Promise<void>
   /** sleep time in ms between reads when at the end of the category */
   tailSleepIntervalMs: number
+  /** sleep time in ms between checkpoint commits */
```
bartelink, line 24: to avoid having two timespans beside each other, maybe put generic stuff like tailsleep and maxreadahead together

then checkpoint/group stuff

then dispatch/scheduling stuff such as maxConcurrentStreams

L26 is it still p-limit

packages/propeller/src/FeedSource.mts
```diff
@@ -2,32 +2,93 @@ import { Batch, Sink } from "./Types.js"
...
   ) {}
 
-  async start(trancheId: string, signal: AbortSignal) {
-    let pos = await this.checkpoints.load(this.groupName, trancheId, this.establishOrigin)
+  async start(signal: AbortSignal) {
+    while (!signal.aborted) {
+      await sleep(this.intervalMs, signal)
+      if (this.committedPos !== this.validatedPos) {
+        await this.checkpoints.commit(this.groupName, this.trancheId, this.validatedPos)
+        this.validatedPos = this.committedPos
```
bartelink: should this be the other way round ?
maybe one liners as to what these mean might help?
maybe call it writtenPos vs currentPos/confirmedPos ?
(in the .fs version, the committed is the value that's actually been written, which is probably a little confusing)
```
        log.Information("Ingester {partition} Ahead {activeReads}/{maxReads} Committed {committed} @ {validated}/{pos} ok {commits} failed {fails} Ingested {batches} ({streams:n0}s {events:n0}e) Cycles {cycles}",
                        partitionId, activeReads, maxReads, Option.toNullable committedEpoch, Option.toNullable validatedEpoch, Option.toNullable readEpoch, commits, commitFails, batchesPended, streamsPended, eventsPended, cycles)
```

packages/propeller/src/FeedSource.mts
```diff
@@ -2,32 +2,93 @@ import { Batch, Sink } from "./Types.js"
...
+}
+
+export type TailingFeedSourceOptions = {
+  tailSleepIntervalMs: number
+  checkpointIntervalMs: number
+  groupName: string
+  checkpoints: ICheckpoints
+  sink: Sink
+  crawl: (trancheId: string, position: bigint, signal: AbortSignal) => AsyncIterable<Batch>
+  establishOrigin?: (tranche: string) => Promise<bigint>
```
bartelink, line 42: would be tempted to put this first too to reflect that it dictates where reading starts

this leaves the later args to group according to the logical flow

or you can put the sink first and have the reverse order too I guess

packages/propeller/src/StreamsSink.mts
```diff
@@ -16,38 +16,116 @@ function groupByStream(items: [StreamName, ITimelineEvent][]): Map<StreamName, I
...
+    return value
+  }
+  return current
+}
+
 export class StreamsSink implements Sink {
-  private readonly limit: LimitFunction
+  private queue = new Queue<Stream>()
+  private streams = new Map<StreamName, Stream>()
+  private batchStreams = new Map<IngesterBatch, Set<StreamName>>()
```
bartelink: these ultimately need to become pairs of stream+position

you want to be able to have a handler respond to a dispatch of N events on a stream by saying it did some, none or all of them (or even set a position beyond where reading was)

in a case where there is a dominant long stream that takes a long time to drain and/or never gets drained, you need to be able to keep declaring earlier batches closed per what the handler has achieved in order to unlock read ahead capacity

and/or allow stuff that's been read be dumped if the handler determines a consumer is way ahead

this becomes particularly relevant if there's streams that take inordinately long to process relative to others (could be rate limiting, could be some wierd data distribution)

packages/message-db/message-db-consumer/src/lib/MessageDbSource.mts
nordfjord: I'll take a look at a better config structure after merging this 👍

PR #34, opened 2023-09-21
add propeller docs

nordfjord:


docs/docs/propeller/index.md
```diff
@@ -0,0 +1,36 @@
...
+pipelines catering specifically for event sourced reactions from Equinox's
+stores.
+
+## StreamsSink
+
+The `StreamsSink` receives batches of events from a source and intelligently
+manages handling them.
+
+- **Stream management**: maintains an ordered queue of incoming streams, ensuring
+  that events are processed in an orderly manner.
```
bartelink, line 18: Probably merits an individual point that the goal is to close batches and checkpoint them, and that this is achieved without needing to stop the world for the checkpointing operation to complete before we move to the next one.

```suggestion
  that reactions processing aligns with the ordering of the triggering events; the aim being to be able to checkpoint attained progress ('tick off batches') as soon as possible.
```

docs/docs/propeller/index.md
```diff
@@ -0,0 +1,36 @@
...
+- **Concurrency control**: To ensure optimal performance, StreamsSink restricts the
+  number of streams and event batches that can be processed at once. It
+  efficiently handles the inflow of new event batches, ensuring the system
+  doesn't get overwhelmed.
+- **Batch merging**: intelligently merges new events with existing streams. This
+  allows for a more streamlined event processing, ensuring that events from
+  multiple batches related to the same stream are handled together.
```
bartelink, line 25: >This
  allows for a more streamlined event processing, ensuring that events from
  multiple batches related to the same stream are handled together.
  
Where multiple batches include events from the same stream, these are supplied to the handler as a single sequence; this allows handler throughput to to be optimized (note the sink also guarantees that there will never be more than one handler processing events for a given stream)

(and if that last part is not true yet, it needs to be; probably worth making it a separate point so its easier to read and stands out)

docs/docs/propeller/index.md
```diff
@@ -0,0 +1,36 @@
...
+
+## TailingFeedSource
+
+`TailingFeedSource` continuously crawls a feed, submitting batches of events to
+a `Sink`.
+
+- **Crawling**: Continuously crawls a given feed. It pauses momentarily between crawls once
```
bartelink, line 32:
```suggestion
- **Crawling**: Continuously crawls a given feed forward from a checkpointed position (or the start of the feed if no batches have yet completed processing). It pauses momentarily between crawls once
```

docs/docs/propeller/index.md
```diff
@@ -0,0 +1,36 @@
...
+  it's reached the tail of the feed.
+- **Checkpointing**: Regularly updates the checkpoint position to match the latest
+  processed batch. Flushes the checkpoint when interrupted to minimise reduntant
```
bartelink, line 35:
```suggestion
  processed batch. Flushes the checkpoint when interrupted to minimise redundant
```

docs/docs/propeller/index.md
```diff
@@ -0,0 +1,36 @@
...
+  multiple batches related to the same stream are handled together.
+
+## TailingFeedSource
+
+`TailingFeedSource` continuously crawls a feed, submitting batches of events to
+a `Sink`.
+
+- **Crawling**: Continuously crawls a given feed. It pauses momentarily between crawls once
+  it's reached the tail of the feed.
+- **Checkpointing**: Regularly updates the checkpoint position to match the latest
```
bartelink, line 34:
```suggestion
- **Checkpointing**: Regularly updates the checkpoint position to reflect the latest completed
```
should mention something about this being async to the processing and deduplicated if there are lots of batches being processed very quickly

docs/docs/propeller/streams-sink.md
```diff
@@ -0,0 +1,37 @@
+# StreamsSink Architecture
+
+## AsyncQueue
+
+The StreamsSink uses an AsyncQueue that provides a node-friendly way to handle a
```
bartelink, line 5:
```suggestion
The `StreamsSink` uses an `AsyncQueue` that provides a Node-friendly way to handle a
```

docs/docs/propeller/streams-sink.md
```diff
@@ -0,0 +1,37 @@
...
+the queue is empty, it waits asynchronously until an item is available or until
+an abort signal is triggered, which would stop the wait and reject the promise
+
+## Receiving batches
+
+For every event within a batch, the StreamsSink checks if its associated Stream
+is already in the processing queue. If it is, the event is combined with it,
+ensuring it's processed alongside previously queued events.
+
+The set of streams associated with the batch are stored in a `Set` so we can
```
bartelink, line 21: set used twice
Per batch, we establish a Set of the streams for which there were events. Whenever a handler completes processing work for a stream, it gets ticked off; the last one results in flagging the batch complete, which the FeedSource then uses to a) update the checkpoint to record the progress attained b) trigger commencement of reading of a replacement batch per the max read ahead policy

docs/docs/propeller/streams-sink.md
```diff
@@ -0,0 +1,37 @@
...
+know when a batch has finished processing and notify the submitter.
+
+The StreamsSink limits the number of in-flight batches to a configured number.
+When a batch is submitted we check the number of in-flight batches and if it has
```
bartelink, line 25:
```suggestion
When a batch is submitted we check the number of in-flight batches and, if it has reached the threshold,
```

docs/docs/propeller/streams-sink.md
```diff
@@ -0,0 +1,37 @@
...
+reached the threshold delays the introduction of the batch's events until a
```
bartelink, line 26: >delays the introduction of the batch's events
wrong semantic - if the work has been done to read, it should be fed into the buffer immediately as the memory is being consumed
it needs to wait _after_ ingesting the batch, allowing completion of a batch to open up capacity to read mode

docs/docs/propeller/streams-sink.md
```diff
@@ -0,0 +1,37 @@
...
+previous batch completes. 
+
+## Worker loops
+
+A set number of worker loops are started based on the configured concurrency
+limit, which then pick up and process streams from the AsyncQueue. Once a stream
+is processed some bookkeeping is required.
+
+1. The stream is removed from each batch's `Set` of streams
+2. If that brings the batch's `Set` size to 0 we know the batch is fully handled
```
bartelink, line 36:
```suggestion
2. If the stream removed was the final one, we declare it complete by calling its associated `onComplete` handler.
```
maybe  "(which wakes anyone waiting for concurrent batching capacity and advances the checkpointable feed position)

docs/docs/reactions/index.md
```diff
@@ -106,6 +106,8 @@ const source = MessageDbSource.create({
   tailSleepIntervalMs: 100, 
   // How many streams are we OK to process concurrently?
   maxConcurrentStreams: 10, 
+  // How many batches can be processod concurrently, checkpointing will always happen in-order
```
bartelink, line 109:
```suggestion
  // How many batches can be pre-loaded beyond what's been completed (checkpointing is always sequential in order of age)
```

docs/docs/reactions/index.md
```diff
@@ -106,6 +106,8 @@ const source = MessageDbSource.create({
...
+  maxConcurrentStreams: 10,
```
bartelink, line 110:
```suggestion
  maxConcurrentBatches: 10, 
```
(but, as fiscussed, probably should stay maxReadAhead)

docs/docs/reactions/projections.md
```diff
@@ -67,35 +64,35 @@ async function handler(streamName: string, events: ITimelineEvent[]) {
   if (statements.length === 0) return
   const client = await pool.connect()
   try {
-    // CAUTION: do not lower the isolation level unless you know what you're doing 
-    await client.query('BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE')
+    // CAUTION: do not lower the isolation level unless you know what you're doing
```
bartelink, line 67:
```suggestion
    // CAUTION: do not lower the isolation level unless you are Kimi Raikkonen
```

docs/docs/reactions/projections.md
```diff
@@ -67,35 +64,35 @@ async function handler(streamName: string, events: ITimelineEvent[]) {
...
-  checkpoints, 
+  categories: [Payer.CATEGORY],
+  groupName: "PayerListModel",
+  checkpoints,
   handler,
-  tailSleepIntervalMs: 5000, 
-  maxConcurrentStreams: 10, 
+  tailSleepIntervalMs: 5000,
+  maxConcurrentStreams: 10,
+  maxConcurrentBatches: 10,
```
bartelink, line 89: maybe start using a diff value for this - 2 or 5 might be good candidates
a) >1 so nobody is waiting for a read
b) not too high so no memory wasted
c) different to the 10 for streams

ultimately if you want to have a lot of stuff buffered, the batch size and the maxReadAhead multipled together are the real number (multiplied by the number of tranches that this node is handling)

if you start heading to 10, you probably just want to up the batch size instead. Downsides:
- bigger batch sizes = less frequent checkpointing -> more rework if killed; less insight into gap/lag from an external observer -> more chunky read traffic -> potentially worse min sunny day reaction latency
- smaller batch sizes -> more memory usage, more store roundtrips, better sunny day reaction latency

The key thing allowing more read ahead can enable is to allow processing to still be reacting to new stuff despite something big/slow/randomly failing in the oldest batch holding up processing in general

This is probably worth calling out a a higher level; it all depends on whether the work being performed by the handler is heterogenous in terms of:
- mix of IO vs compute bound
- mix of high/low traffic streams
- ability of a handler to handle sequences of events efficiently relative to processing them individually

docs/docs/reactions/projections.md
nordfjord: Yes!

docs/docs/reactions/index.md
nordfjord: No reason to break with propulsion namings I guess. 'tis been undone

docs/docs/reactions/projections.md
nordfjord: Great point, I'll add a section on this and change the default in the docs to 3

nordfjord: See `configuring.md`

docs/docs/propeller/configuring.md
```diff
@@ -0,0 +1,50 @@
+# Configuring
+
+## Controlling throughput 
+
+Propeller has three parameters that work together to control throughput,
+`batchSize`, `maxConcurrentStreams` and `maxReadAhead`. The first controls the
```
bartelink, line 6: maybe: they control respectively:

docs/docs/propeller/configuring.md
```diff
@@ -0,0 +1,50 @@
...
+size of batches read from the source, the second how many streams can be
+concurrently processed, and the third how many batches can be buffered in
+memory.
+
+It is generally good to set `maxReadAhead` to at least `2` as that minimises
+downtime by allowing the feed to fetch a batch concurrently with stream
```
bartelink, line 12: It is generally useful to set `maxReadAhead` to at least `2` to minimize
idle time by allowing the feed to fetch the next batch during the handling of the streams. The number of events held in memory will be up to `batchSize *
maxReadAhead` (the last batch may be incomplete).

docs/docs/propeller/configuring.md
```diff
@@ -0,0 +1,50 @@
...
+should consider whether the work being performed by the handler:
+
+- is IO or compute bound
+- handles a mix of high/low traffic streams
+- can handle sequences of events efficiently relative to processing them
+  individually
+
+## Sleeping at the tail 
+
+When the tail (end) of a feed has been reached propeller will sleep for a
```
bartelink, line 36: When the tail (end) of a feed has been reached, Propeller will sleep for a

docs/docs/propeller/configuring.md
```diff
@@ -0,0 +1,50 @@
...
+is required. A lower value will mean more frequent round-trips to check for data
+and as such should be considered in terms of RUs and DB load.
+
+## Checkpointing
+
+Checkpointing happens asynchronously on a configurable interval. Propeller will
+additionally attempt to flush a checkpoint once its AbortSignal is fired. Note
+that idempotent handling is a requirement, checkpointing is an optimisation and
+your system should behave the same whether it starts from scratch on every start
+or at the last stored checkpointed.
```
bartelink, line 49: extend this to mention that rewinding the checkpoint to zero for a full replay should be considered a completely normal thing. Don't want the uninitiated walking away thinking the point is that every 500 starts there might be an event or two repeated here and there, i.e. that this is just some tinfoil hat corner case.

docs/docs/propeller/index.md
```diff
@@ -0,0 +1,40 @@
+---
+sidebar_position: 4
+---
+
+# Propeller
+
+Propeller is EquinoxJS's "[propulsion](https://github.com/jet/propulsion)
+lite". It's a set of abstractions for building reactive event processing
```
bartelink, line 8: the abstractions are not the sell really - its the fact that there are impls, and that they have a ecent structure that will yield you a testable/separated handler not mangled in store specific bpoilerplate.

It's a set of components for building ...
The primary purpose is to manage reactions to events emanating from an event store's feed.
The components all follow common high level patterns, with specific treatments and options to allow you to manage specific details of the actual store you're using.
The high level structure is also designed to work well in terms of being able to test handlers in isolation, or deal with similar needs, such as continually ingesting data from third party event feeds.

(last point is probably not intro material)

docs/docs/propeller/index.md
```diff
@@ -0,0 +1,40 @@
...
+
+The `StreamsSink` receives batches of events from a source and intelligently
+manages handling them.
+
+- **Stream management**: maintains an ordered queue of incoming streams,
+  ensuring that reactions processing aligns with the ordering of the triggering
+  events; the aim being to be able to checkpoint attained progress ('tick off
+  batches') as soon as possible.
+- **Concurrent stream handling**: To ensure optimal performance, `StreamsSink`
+  restricts the number of streams that are processing at any given moment to a
```
bartelink, line 22: ->that are processing
for which handlers are processing

maybe 'in flight' ?

docs/docs/propeller/index.md
```diff
@@ -0,0 +1,40 @@
...
+  configurable limit. Additionally guarantees that each stream can have at most
```
bartelink, line 23: This is not a lil bonus - its a central thing

>Additionally guarantees that each stream can have at most
  one handler actively processing.

Note, while handlers work concurrently, there's only ever a single handler in flight for any given stream; it gets passed the queue of all pending events for that stream 

(cant think of a good sentence to convey "and you can do them sequentially, run them in parallel, do batch SQL", without contention knowing that there are not other events for that stream being handled at the same time)

docs/docs/propeller/index.md
```diff
@@ -0,0 +1,40 @@
...
+  one handler actively processing.
+- **Batch merging**: intelligently merges new events with existing streams.
```
bartelink, line 25: maybe this goes before the preceding point, leading into the fact that the handler gets the latest coalesced stream events from across batches

docs/docs/propeller/index.md
```diff
@@ -0,0 +1,40 @@
...
+## TailingFeedSource
+
+`TailingFeedSource` continuously crawls a feed, submitting batches of events to
+a `Sink`.
+
+- **Crawling**: Continuously crawls a given feed forward from a checkpointed
+  position (or the start of the feed if no batches have yet completed
+  processing). It pauses momentarily between crawls once it's reached the tail
+  of the feed.
+- **Checkpointing**: Regularly updates the checkpoint position to reflect the
```
bartelink, line 38: regularly sounds like it's a pretty random sort of a contraption. Periodic?

docs/docs/reactions/index.md
```diff
@@ -106,6 +106,8 @@ const source = MessageDbSource.create({
   tailSleepIntervalMs: 100, 
   // How many streams are we OK to process concurrently?
   maxConcurrentStreams: 10, 
+  // How many batches can be pre-loaded beyond what's been completed (checkpointing is always sequential in order of age)
```
bartelink, line 109: need to convey that it includes the actual batch being processed. the concurrent batches name was perhaps better!

docs/docs/reactions/projections.md
```diff
@@ -67,35 +64,35 @@ async function handler(streamName: string, events: ITimelineEvent[]) {
   if (statements.length === 0) return
   const client = await pool.connect()
   try {
-    // CAUTION: do not lower the isolation level unless you know what you're doing 
-    await client.query('BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE')
+    // CAUTION: do not lower the isolation level unless you are Kimi Raikkonen
```
bartelink, line 67: I guess if you know you know but https://www.google.com/search?q=raikkonen+quote gives a good result at this moment in time for me!

packages/message-db/message-db-consumer/src/lib/MessageDbSource.mts
```diff
@@ -26,7 +26,7 @@ interface CreateOptions {
   /** The maximum number of concurrent streams to process */
   maxConcurrentStreams: number
   /** The maximum number of batches in-flight */
```
bartelink, line 28: def: 10 (should it be?()

docs/docs/propeller/index.md
bartelink: maybe it should allude to or mention that the wait to checkpoint is not part of the processing time overall and/or that it dedups checkpoint writes if there are lots in a replay scenario?

docs/docs/propeller/streams-sink.md
```diff
@@ -0,0 +1,38 @@
...
+For every event within a batch, the StreamsSink checks if its associated Stream
+is already in the processing queue. If it is, the event is combined with it,
+ensuring it's processed alongside previously queued events.
+
+The streams associated with the batch are stored in a `Set` so we can know when
+a batch has finished processing and notify the submitter.
+
+The StreamsSink limits the number of in-flight batches to a configured number.
+After a batch is submitted we check the number of in-flight batches and, if it
+has reached the threshold, delays the introduction of the next batch until a
```
bartelink, line 26: delays fetching until capacity is available?

docs/docs/propeller/streams-sink.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Worker loops
+
+A set number of worker loops are started based on the configured concurrency
+limit, which then pick up and process streams from the AsyncQueue. Once a stream
+is processed some bookkeeping is required.
+
+1. The stream is removed from each batch's `Set` of streams
+2. If the stream removed was the final one, we declare it complete by calling
+   its associated `onComplete` handler. and call its associated `onComplete`
```
bartelink, line 37: mangled - maybe ref that this lets the caller invalidate its record of the checkpoint pos

docs/docs/reactions/projections.md
nordfjord: image

docs/docs/propeller/configuring.md
nordfjord: Addressed in #37

docs/docs/propeller/configuring.md
nordfjord: addressed in #37

docs/docs/propeller/index.md
nordfjord: You are absolutely correct. It's just so much easier to write about the abstractions :(

PR #35, opened 2023-09-21
improve streams sink

nordfjord:


packages/dynamo-store/source/src/DynamoStoreSource.mts
```diff
@@ -305,14 +305,14 @@ export class DynamoStoreSource {
...
 
   constructor(
-    private readonly index: AppendsIndex.Reader,
-    private epochs: AppendsEpoch.Reader.Service,
+    index: AppendsIndex.Reader,
+    epochs: AppendsEpoch.Reader.Service,
     private readonly options: Omit<CreateOptions, "context">,
   ) {
     if (!this.options.categories && !this.options.streamFilter) {
       throw new Error("Either categories or categoryFilter must be specified")
```
bartelink, line 313:
```suggestion
      throw new Error("Either categories or streamFilter must be specified")
```

packages/propeller/src/StreamsSink.mts
```diff
@@ -40,59 +106,52 @@ export class StreamsSink implements Sink {
     Object.assign(this.tracingAttrs, attrs)
   }
 
-  start(signal: AbortSignal) {
+  private handleStreamCompletion(stream: Stream) {
+    for (const [batch, streams] of this.batchStreams) {
+      streams.delete(stream)
+      if (streams.size === 0) {
+        batch.onComplete()
+        this.batchStreams.delete(batch)
```
bartelink, line 114: should always be head?
if not, you might be notifying in wrong order

packages/propeller/src/StreamsSink.mts
```diff
@@ -40,59 +106,52 @@ export class StreamsSink implements Sink {
...
-      this.onReady = () => {
-        signal.removeEventListener("abort", abort)
-        resolve()
-      }
-    })
+    for (let i = 0; i < this.maxConcurrentStreams; ++i) {
+      workers[i] = new QueueWorker(
+        async (signal) => {
+          const stream = await this.queue.tryGetAsync(signal)
+          this.streams.delete(stream.name)
```
bartelink, line 135: just checking you have a way for workers to guarantee that they don't pick up a stream while another 'owns' it, e.g. if a fresh batch has just been added

nordfjord: Yeah, this is JS. All the work is happening on a single thread, so the removal from the map is enough to ensure it won’t be modified while handled

packages/propeller/src/StreamsSink.mts
nordfjord: Streams are processed in-order, I guess it’s possible if you’re at the tail that your processing of the last stream of batch n-1 takes longer than processing the single event in batch n, which would trigger the complete out of order. I’ll add some defensiveness

packages/propeller/src/StreamsSink.mts
bartelink: OK, but not seeing how this scenario is covered:

- B1 has S1 and S2
- B2 has S1 and S3
- there are two handlers, H1, H2

B2 arrives after H1 and H2 have started on S1 and S2 respectively

After H1 finishes S2, how does it know that H1 is doing stuff on S1 so it can only work on S3?

nordfjord: You’re right, that’s not catered for, and needs to be 😬

bartelink: Having long lived stream state to facilitate this negotiation also means you'll be able to record an attained position per stream.

You'll probably independently record the high water mark per stream at the batch level too though:
- so when H1 has completed S1 you can decide whether both B1 and B2 have had their S1 events covered
- so handers can return a result that conveys that (even though only the ones from B1 were supplied) the S1 items in B2 should be skipped as the reactor/read model is at position X
- this also lets the handler say "OK you sent me S1 event 0 - I went to read it but didnt see it. Put it back in the queue please". This allows newer items to join for the next attempt. It also allows you to implement backoffs in the face of rate limiting or similar a la https://github.com/jet/propulsion/pull/129 (also useful for e.g. the shipping watchdog where you're running a process that's responsible for tracking workflow timeouts by cancelling them - the current impl does a blocking sleep)

See also https://github.com/jet/propulsion/blob/8eef5211e398a8243916c967096b64976ef112c9/src/Propulsion/Streams.fs#L543-L568

And a per stream pos can also help with implementing gap detection lol

PR #38, opened 2023-09-24
docs: service anatomy

nordfjord:


docs/docs/anatomy.md
```diff
@@ -0,0 +1,254 @@
+---
+sidebar_position: 3
+---
+
+# Anatomy of an Equinox Service
+
+An Equinox service is composed of a few modules (namespaces).
+
+- **Stream**: tells us which category the service writes to and how the identity
+  of the stream is composed
```
bartelink, line 10:
```suggestion
  of the stream is composed (and the reverse operations for when running Reaction logic)
```

docs/docs/anatomy.md
```diff
@@ -0,0 +1,254 @@
...
+- **Events**: tells us which events the service writes
+- **Fold**: tells us how the events are folded into state
+- **Decide**: tells us which actions can be taken upon the state (resulting in
+  new events being written)
+- **Service**: the class that wraps the above into a cohesive domain service.
```
bartelink, line 15: maybe a `Queries` section for rendering helpers and dto types like https://github.com/jet/equinox/pull/429 or is this too early?

nordfjord: Definitely not too early, I'll add it and mention it as optional

docs/docs/anatomy.md
```diff
@@ -0,0 +1,254 @@
...
+- **Config**: tells us how to wire the service against concrete stores,
```
bartelink, line 16:
```suggestion
- **Config**: shows how we've opted to to bind the Service to Streams in a Concrete Store (which Access Strategies to apply, if any)
```

docs/docs/anatomy.md
```diff
@@ -0,0 +1,254 @@
...
+  including important information like access strategies.
+
+Let's see what each of these modules looks like for a simple service for
+checking in and out of an appointment.
```
bartelink, line 20:
```suggestion
checking in and out for an Appointment Booking.
```

docs/docs/anatomy.md
```diff
@@ -0,0 +1,254 @@
...
+sidebar_position: 3
+---
+
+# Anatomy of an Equinox Service
+
+An Equinox service is composed of a few modules (namespaces).
+
+- **Stream**: tells us which category the service writes to and how the identity
+  of the stream is composed (and the reverse operations for when running Reaction logic)
+- **Events**: tells us which events the service writes
```
bartelink, line 11: and how they are to be mapped to/from the stored form (e.g. applying wrapper types over primitives)

but only if you an convey it more tersely
the aim being to sell why they are needed and not overkill in a one liner

docs/docs/anatomy.md
```diff
@@ -0,0 +1,254 @@
...
+      case "CheckedOut":
+        return { ...state, checkedOut: event.data.timestamp }
+      case "ActualsOverridden":
+        return event.data
+    }
+  }
+}
+```
+
+## The Decide
```
bartelink, line 79: this reads wacko
And while its perfect for a small example, it odoes not show how you might subdivide it for multiple decisions

docs/docs/anatomy.md
```diff
@@ -0,0 +1,254 @@
...
+
+```ts
+export namespace Decide {
+  import Event = Events.Event
+  import State = Fold.State
+  export const checkIn =
+    (timestamp: Date) =>
+    (state: State): Event[] => {
+      if (state.checkedIn) {
```
bartelink, line 88: while this is subjective and has pros
I think 
1. doing a check on the conjunction and throwing
2. then having an if state.checkedIn vs not to illustrate silent idempotent handling as the main thing
3. then the happy path
might convey the rules of the game better (also `no change` conveyed it to me, but its a bit subtle and might merit something more long winded like saying 'caller is expected to guard against this' and `we accept retried requests (with identical timestamps)`)

docs/docs/anatomy.md
```diff
@@ -0,0 +1,254 @@
...
+      if (state.checkedOut) {
+        // no change
+        if (+state.checkedOut === +timestamp) return []
+        throw new Error("Already checked out with different timestamp")
+      }
+      return [{ type: "CheckedOut", data: { timestamp } }]
+    }
+
+  export const manuallyOverride = (checkedIn: Date, checkedOut: Date) => (state: State) => {
+    if (+checkedIn === +state?.checkedIn && +checkedOut === +state?.checkedIn) return []
```
bartelink, line 107: if there are comments on the above explicatting the concept, this is a fantastic example

docs/docs/anatomy.md
bartelink: also maybe call out that the other things are private/pubternal and this is the thing that you talk to

docs/docs/anatomy.md
```diff
@@ -0,0 +1,254 @@
...
+  ) {}
+
+  checkIn(appointmentId: AppointmentId, userId: UserId, timestamp: Date) {
+    const decider = this.resolve(appointmentId, userId)
+    return decider.transact(Decide.checkIn(timestamp), LoadOption.AssumeEmpty)
+  }
+
+  checkOut(appointmentId: AppointmentId, userId: UserId, timestamp: Date) {
+    const decider = this.resolve(appointmentId, userId)
+    return decider.transact(Decide.checkOut(timestamp), LoadOption.AssumeEmpty)
```
bartelink, line 131: dont assume empty for checkOut
a) coz it wont be
b) to underscore that most of the time you dont do or think about fancy options

nordfjord:
```suggestion
    return decider.transact(Decide.checkOut(timestamp))
```

docs/docs/anatomy.md
```diff
@@ -0,0 +1,254 @@
...
+
+export type Config =
+  | { store: Store.Memory; context: VolatileStore<string> }
+  | { store: Store.MessageDb; context: MessageDB.MessageDbContext; cache: ICache }
+  | { store: Store.Dynamo; context: DynamoDB.DynamoStoreContext; cache: ICache }
+
+// prettier-ignore
+export namespace MessageDb {
+  import AccessStrategy = MessageDB.AccessStrategy
+  import MessageDbCategory = MessageDB.MessageDbCategory
```
bartelink, line 171: short nams like L195?

docs/docs/anatomy.md
```diff
@@ -0,0 +1,254 @@
...
+  export function createCached<E, S, C>(name: string, codec_: ICodec<E, string, C>, fold: (s: S, e: E[]) => S, initial: S, access: AccessStrategy<E, S>, { context, cache }: Config) {
+    const caching = CachingStrategy.Cache(cache)
+    const codec = Codec.compress(codec_)
+    return Category.create(context, name, codec, fold, initial, caching, access);
+  }
+  export function createUnoptimized<E, S, C>(name: string, codec: ICodec<E, string, C>, fold: (s: S, e: E[]) => S, initial: S, config: Config) {
+    const access = AccessStrategy.Unoptimized()
+    return Dynamo.createCached(name, codec, fold, initial, access, config)
+  }
+  export function createLatestKnown<E,S,C>(name: string, codec: ICodec<E, string, C>, fold: (s: S, e: E[]) => S, initial: S, config: Config) {
```
bartelink, line 207: this is last in mdb - prob last here too as pretty rare

docs/docs/anatomy.md
nordfjord: > pubternal

Only to be discussed at the pub, never in the office

docs/docs/anatomy.md
```diff
@@ -0,0 +1,254 @@
...
+```ts
+import * as Config from "../equinox-config"
+class Service {
+  // ... as above
+
+  // prettier-ignore
+  static resolveCategory(config: Config.Config) {
+    switch (config.store) {
+      case Config.Store.Memory:
+        return Config.MemoryStore.create(Stream.name, Events.codec, Fold.fold, Fold.initial, config)
```
bartelink, line 240:
```suggestion
        return Config.MemoryStore.create(Stream.category, Events.codec, Fold.fold, Fold.initial, config)
```

docs/docs/anatomy.md
```diff
@@ -0,0 +1,254 @@
...
+      case Config.Store.MessageDB:
+        return Config.MessageDB.createUnoptimized(Stream.name, Events.codec, Fold.fold, Fold.initial, config)
+      case Config.Store.Dynamo:
+        return Config.Dynamo.createUnoptimized(Stream.name, Events.codec, Fold.fold, Fold.initial, config)
+    }
+  }
+  static create(config: Config.Config) {
+    const category = Service.resolveCategory(config)
+    const resolve = (appointmentId: AppointmentId, userId: UserId) =>
+      Decider.forStream(category, Stream.streamId(appointmentId, userId), null)
```
bartelink, line 250: name the null and/or leave it optional?

nordfjord: ATM the null is necessary, will be making it optional soon 👍

docs/docs/anatomy.md
bartelink: yes, think so

docs/docs/anatomy.md
```diff
@@ -237,7 +237,7 @@ class Service {
   static resolveCategory(config: Config.Config) {
     switch (config.store) {
       case Config.Store.Memory:
-        return Config.MemoryStore.create(Stream.name, Events.codec, Fold.fold, Fold.initial, config)
+        return Config.MemoryStore.create(Stream.category, Events.codec, Fold.fold, Fold.initial, config)
```
bartelink, line 240: see clones

docs/docs/anatomy.md
```diff
@@ -6,19 +6,30 @@ sidebar_position: 3
...
+- **Decide**: shows which actions can be taken upon the state (resulting in
   new events being written)
 - **Service**: the class that wraps the above into a cohesive domain service.
 - **Config**: shows how we've opted to to bind the Service to Streams in a Concrete Store (which Access Strategies to apply, if any)
   including important information like access strategies.
 
+:::tip
+
+Everything above `Service` is considered internal to the module. It is exported
+as a conveninence for testing, but other modules should not take a dependency on
```
bartelink, line 22:
```suggestion
as a convenience for testing, but other modules should not take a dependency on
```

docs/docs/anatomy.md
```diff
@@ -148,6 +163,8 @@ export class Service {
 <details>
 <summary>See the `../equinox-config.ts` snippet</summary>
 
+Feel free to copy-paste this snippet for your own project.
```
bartelink, line 166: Maybe mention that in general you have one set of this binding per repo ?
And that in most real apps you'll typically have bindings for one concrete store and the memory store

docs/docs/anatomy.md
```diff
@@ -168,22 +185,23 @@ export type Config =
...
     const access = AccessStrategy.Unoptimized<E, S>()
     return MessageDb.createCached(name, codec, fold, initial, access, config)
   }
-  export function createSnapshotted<E, S, C>(name: string, codec: ICodec<E, string, C>, fold: (s: S, e: E[]) => S, initial: S, eventName: string, toSnapshot: (s: S) => E, config: { context: MessageDbContext; cache: ICache }) {
+  export function createSnapshotted<E, S, C>(name: string, codec: ICodec<E, string, C>, fold: (s: S, e: E[]) => S, initial: S, eventName: string, toSnapshot: (s: S) => E, config: Config) {
     const access = AccessStrategy.AdjacentSnapshots(eventName, toSnapshot)
     return MessageDb.createCached(name, codec, fold, initial, access, config)
   }
-  export function createLatestKnown<E, S, C>(name: string, codec: ICodec<E, string, C>, fold: (s: S, e: E[]) => S, initial: S, config: { context: MessageDbContext; cache: ICache }) {
+  export function createLatestKnown<E, S, C>(name: string, codec: ICodec<E, string, C>, fold: (s: S, e: E[]) => S, initial: S, config: Config) {
```
bartelink, line 204: latesknown is not last for Dynamo

docs/docs/anatomy.md
```diff
@@ -168,22 +185,23 @@ export type Config =
 // prettier-ignore
 export namespace MessageDb {
   import AccessStrategy = MessageDB.AccessStrategy
```
bartelink, line 187: hm should AccessStrategy be prefixed too

nordfjord: I'd rather the other stuff lost its prefix

```ts
DynamoStore.DynamoStoreCategory.create(....)
// vs
DynamoStore.Category.create(...)
```

bartelink: yeah, but some names are very generic (client, category, context)
and then there are other conventional names like Config, Store, Streams also in the same scope
Given the names should only really be used in the context of one of those Config binding wiring blocks, the fact they are a bit long is no real harm
Also Category would clash with Equinox.Category (and I wanted to rename ISyncContext to Context but again same concerns stop me esp given how overloaded the term Context is)

A lot of these things get close enough to files that have domain logic too - things like Category etc can easily encroach on and/or require constant disambiguation.

Can't see where I saw the argument for the long names - it was in some Azure Core guidlines or sometihng (not they should be conodered arbiters of taste for even 5 mins!)

docs/docs/anatomy.md
nordfjord: Yeah, definitely more common to use snapshots than latest event in dynamo. In MDB I've got more services on LatestKnownEvent than AdjacentSnapshots.

docs/docs/anatomy.md
nordfjord: It is now

docs/docs/anatomy.md
nordfjord: modified

docs/docs/anatomy.md
nordfjord: This'll be the second item in the nav. Not sure if it's the right place for a full-blown complex thing. Don't want to scare people just yet 😅

docs/docs/anatomy.md
nordfjord: Query section added

docs/docs/anatomy.md
nordfjord: see #41

PR #40, opened 2023-09-25
fix!: upgrade node to v20, fix abort handling

nordfjord:


packages/propeller/src/Queue.ts
```diff
@@ -88,7 +88,7 @@ export class AsyncQueue<T> {
     const value = this.queue.tryFind(predicate)
     if (value) return value
     return new Promise<T>((resolve, reject) => {
-      const abort = () => reject(new Error("Aborted"))
+      const abort = () => reject(new DOMException("The operation was aborted", "AbortError"))
```
sroze, line 91: @nordfjord why `DOMException` in this very backend-y package? It means it doesn't work with Node < 20, either.

sroze: image

nordfjord: Because that's what node throws from `AbortController#abort`

nordfjord: Would be even worse to have multiple types of exceptions representing abort errors IMO

nordfjord: See https://github.com/nodejs/node/blob/main/lib/internal/abort_controller.js#L178-L179

PR #42, opened 2023-10-06
fix: Bubble up database errors in `message-db` component

sroze:
Seen by using `equinox-js` in practice: when we didn't configure the database connection details properly, the only thing that application engineer sees is:

> Error: Concurrency violation; aborting after 3 attempts.

Which is particularly unhelpful. We need to find a way to bubble up the database error, too.


nordfjord:
Closing as superseded by #45

PR #45, opened 2023-10-06
fix: catch the wrong expected version error from MDB

nordfjord:


sroze:
I think we should add tests :)

PR #57, opened 2023-11-13
fix: use Promise.allSettled

nordfjord:
before this change aborting the signal passed to FeedSource would
trigger an immediate rejection of the sink promise while the source
promise would do some processing. As such Promise.all would reject
immediately, meaning we couldn't effectively wait for the post
processing of the source

### DIFF
--- a/packages/dynamo-store/source/src/DynamoStoreSource.test.mts
+++ b/packages/dynamo-store/source/src/DynamoStoreSource.test.mts
@@ -41,6 +41,10 @@ const waiter = () => {
   ] as const
 }
 
+const throwIfActive = (signal: AbortSignal) => (e: any) => {
+  if (!signal.aborted) throw e
+}
+
 test("Correctly batches stream handling", async () => {
   const store = new VolatileStore()
   const index = AppendsIndex.Reader.createMem(store)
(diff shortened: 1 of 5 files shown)


packages/dynamo-store/source/src/DynamoStoreSource.test.mts
```diff
@@ -41,6 +41,10 @@ const waiter = () => {
   ] as const
 }
 
+const throwIfActive = (signal: AbortSignal) => (e: any) => {
```
steve-birdie, line 44: `unknown` rather than `any` ?

packages/propeller/src/FeedSource.mts
```diff
@@ -101,18 +101,28 @@ export class TailingFeedSource extends EventEmitter {
           wasTail = batch.isTail
         }
       }
+
+      if (!signal.aborted && checkpointError) {
```
steve-birdie, line 105: Should we be checking the aborted case here - the `checkpointError` should only be set in the event it wasn't aborted at the point the error was caught - in which case should we throw it anyway from here (even if someone has aborted between the error being caught, and this logic firing)?

steve-birdie: Oh sorry - I now see we're setting checkpointError unconditionally - tho should we?

PR #58, opened 2023-11-17
refactor: Codec.tryDecode renamed to decode

nordfjord:
this reduces confusion as `tryDecode` can actually throw an exception
and the undefined return value is to indicate that the event, while
valid (probably at some other time in the past), should be ignored

### DIFF
--- a/docs/docs/core-concepts/codec.md
+++ b/docs/docs/core-concepts/codec.md
@@ -17,7 +17,7 @@ like this:
 
 ```ts
 const codec: Codec<Event, string> = {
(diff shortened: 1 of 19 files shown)


docs/docs/core-concepts/codec.md
```diff
@@ -79,6 +79,13 @@ const codec = Codec.map(
 )
 ```
 
+When upcasting is used the `decode` method of the resulting codec will drop
```
bartelink, line 82:
```suggestion
When upcasting is used, the `decode` method of the resulting codec will drop
```

PR #59, opened 2023-11-29
feat: improve context to allow proper correlation tracking

nordfjord:
This change allows proper tracking of correlation ids for equinox projects. Before this change we couldn't control the ID of the event in `mapMeta` which made for some difficulties initialising the correlation context (the first event in the chain has `id = corrId = causId` so instead of restricting mapMeta to provide meta, we allow it to map both the ID and the meta. (Potential rename to `mapCausation`)

### DIFF
--- a/apps/example/src/domain/invoice-auto-emailer.ts
+++ b/apps/example/src/domain/invoice-auto-emailer.ts
@@ -5,47 +5,56 @@ import z from "zod"
 import { EmailSender, ISendEmails } from "./send-email.js"
 import * as Config from "../config/equinox.js"
 import * as Invoice from "./invoice.js"
+import { Context } from "../context/context.js"
 
-const CATEGORY = "InvoiceAutoEmail"
-const streamId = StreamId.gen(InvoiceId.toString)
+export namespace Stream {
+  export const category = "InvoiceAutoEmail"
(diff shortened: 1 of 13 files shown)


apps/example/src/domain/invoice-auto-emailer.ts
```diff
@@ -5,47 +5,56 @@ import z from "zod"
 import { EmailSender, ISendEmails } from "./send-email.js"
 import * as Config from "../config/equinox.js"
 import * as Invoice from "./invoice.js"
+import { Context } from "../context/context.js"
 
-const CATEGORY = "InvoiceAutoEmail"
-const streamId = StreamId.gen(InvoiceId.toString)
+export namespace Stream {
```
bartelink, line 10: wait what happened to CategoryId?

nordfjord: I was waiting on you to land it in dotnet templates 😊

bartelink: I did and I CC'd you a few days back (though the docs are lagging, as they sometimes do)

nordfjord: Oh, my bad then, I’m on holiday right now and was travelling last week. I’ll get on it 💪

PR #65, opened 2024-01-03
feat: actually validate uuids

nordfjord:
So far the "parse" of the uuid modules has been kind of a lie, it just lowercases whatever you give it and pretends. This change will actually ensure that the Uuid is a hexadecimal string with 0 or 4 dashes.

Curious for opinions from @sroze and @everzet whether this is a good idea

### DIFF
--- a/packages/core/src/lib/Uuid.ts
+++ b/packages/core/src/lib/Uuid.ts
@@ -2,6 +2,33 @@ import { randomUUID } from "crypto"
 
 export type Uuid<T> = string & { __brand: T }
 
+const zero = "0".charCodeAt(0)
+const nine = "9".charCodeAt(0)
+
+const isDigit = (c: number) => zero <= c && c <= nine
+const isHex = (c: number) => isDigit(c) || (97 <= c && c <= 102)
+
+function parse<T>(str: string): Uuid<T> {
+  str = str.toLowerCase()
+  if (str.length !== 36 && str.length !== 32) {
+    throw new Error("Invalid UUID")
+  }
+  let dashCount = 0
+  for (let i = 0; i < str.length; ++i) {
+    if (str[i] === "-") {
+      ++dashCount
+      continue
+    }
+    if (!isHex(str.charCodeAt(i))) {
+      throw new Error("Invalid UUID")
+    }
+  }
+  if (dashCount !== 4 && dashCount !== 0) {
+    throw new Error("Invalid UUID")
+  }
+  return str as Uuid<T>
+}
+
 export type UuidModule<T> = {
   create: () => Uuid<T>
   toString: (uuid: Uuid<T>) => string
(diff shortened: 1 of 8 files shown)


sroze:
Just so I understand what's the consequence on clients using the library? Does it change anything about the API contract or is this just an internal matter?

packages/core/src/lib/Uuid.ts
```diff
@@ -2,6 +2,17 @@ import { randomUUID } from "crypto"
 
 export type Uuid<T> = string & { __brand: T }
 
+// permissive uuid regex that allows non-standard uuids
+const regex = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/
+
+function parse<T>(str: string): Uuid<T> {
+  const lower = str.toLowerCase()
+  if (!regex.test(str)) {
```
bartelink, line 10: Did you intend to test `lower`, or should L9 be inlined?
Could this create function become a gate that canonicalises to 'N` format (i.e. no dashes/braces etc, lowercase) ?

packages/core/src/lib/Uuid.ts
```diff
@@ -2,6 +2,17 @@ import { randomUUID } from "crypto"
 
 export type Uuid<T> = string & { __brand: T }
 
+// permissive uuid regex that allows non-standard uuids
```
bartelink, line 5: doesnt look very permissive from where I'm sitting?

nordfjord: Permissive in that it accepts any hexadecimal string as long as it looks like a uuid. The canonical regex looks like this https://github.com/uuidjs/uuid/blob/main/src/regex.js

packages/core/src/lib/Uuid.ts
```diff
@@ -2,6 +2,17 @@ import { randomUUID } from "crypto"
...
 
+// Custom regex: admits any hex values matching the normal dashed cluster grouping structure
+// loose in that it does not attempt to reject 'impossible/undefined' based on the UUID spec
+const regex = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i
+
+function parse<T>(str: string): Uuid<T> {
+  if (!regex.test(str)) {
+    throw new Error(`Uuid: invalid format '${str}'`)
+  }
+  return str.toLowerCase() as Uuid<T>
```
bartelink, line 13: if you're going to canonicalize as lower case, better to do that immediately and validate that the thing you'll ultimately store is valid/compliant
(and, to be pedantic/emphasize that, I'd make the regeex not be case-insensitive)

nordfjord: I don't really have a preference, so happy to go with that.

PR #70, opened 2024-01-07
feat: add support for immediately consistent projections

nordfjord:
Relates to discussion #69

This is a proof of concept for adding immediately consistent projections to
message-db. For a lot of applications this would provide a sensible way to
create common list projections for the 95% use case at the cost of some write
performance.

As trade-offs go it's certainly not a bad one when contrasted with the
infrastructure necessary to manage eventual consistency for highly interactive
frontends. Instead of implementing the requisite entity-level caching and
optimistic update, end-users could instead read their own writes for the "main"
projection. There might still be other projections and reactions that happen
asynchronously, but for this common use case of "I want to update my entity and
route the user to an already up-to-date list of entities" an immediately
consistent projection seems the better tool for the job.

We'll need to document the feature and the caveats like you need to use a user that has access to the message_store schema and the projection schema, and

1. This should only be used to upsert rows in a secondary table where each row
   maps to a single stream in the category
2. Seriously, don't try to send an SMS or an e-mail in one of these as an
   expedient
3. Read 1 and 2 very carefully
4. Unless you're Kimi Raikkonen, read it again, just in case

Other caveats to mention are around what happens when you get things wrong (i.e. migrations). Since these are only update on write you'd need to devise some script to update the projection (possibly using a reaction)


apps/example/src/config/equinox.ts
```diff
@@ -54,8 +56,9 @@ export namespace MessageDb {
     return MessageDb.createCached(name, events, fold, access, config)
   }
 
-  export function createProjected<E, S, C>(name: string, events: Events<E, C>, fold: Fold<E,S>, config: { context: MessageDbContext; cache: ICache }) {
-    const access = AccessStrategy.AdjacentSnapshots
+  export function createProjected<E, S, C>(name: string, events: Events<E, C>, fold: Fold<E,S>, project: (client: Client, streamId: StreamId, state: S) => Promise<void>, sub: AccessStrategy<E, S> | undefined, config: { context: MessageDbContext; cache: ICache }) {
```
bartelink, line 59: can you mix the project function into each case?
If not maybe make it an optional arg on the mdb category?

nordfjord: That's a good point, I'll give that a try, see if I like the API better that way

bartelink: yeah, but it is largely orthogonal to the access strategy
(but I do get how you want to put there difference in people's faces to make people aware of store/category specific features given that the access strategy is the primary place such features and considerations surface)

nordfjord: done

apps/example/src/read-models/PayerReadModel.ts
```diff
@@ -1,38 +1,36 @@
...
-  if (!event) return []
-  const version = events[events.length - 1]!.index
-  switch (event.type) {
-    case "PayerProfileUpdated":
-      const data = event.data
-      return [Upsert({ id: id, version, name: data.name, email: data.email })]
-    case "PayerDeleted":
-      return [Delete({ id: id, version })]
+  if (!state) return [Delete({ id, version })]
+  return [Upsert({ id, version, name: state.name, email: state.email })]
```
bartelink, line 19: any way for this to be conditional on version > current
would stop time travel under replay and prevent null writes

apps/example/src/read-models/PayerReadModel.ts
```diff
@@ -1,38 +1,36 @@
...
+import { forEntity, Change, createHandler, MinimalClient } from "@equinox-js/projection-pg"
 
-type Payer = { id: PayerId; version: bigint; name: string; email: string }
+type Payer = { id: PayerId; name: string; email: string; version: string }
 
-const { Delete, Upsert } = forEntity<Payer, "id"| "version">()
+const { Delete, Upsert } = forEntity<Payer, "id" | "version">()
 
-export const projection = { table: "payer", id: ["id"], version: "version"}
+export const projection = { table: "payer", id: ["id"], version: "version" }
```
bartelink, line 11: projection vs project - good to remove overloading
any less ambiguous name?

apps/example/src/read-models/PayerReadModel.ts
nordfjord: It already is, that’s what upset does

apps/example/src/read-models/PayerReadModel.ts
```diff
@@ -1,38 +1,36 @@
...
-  const event = Payer.Events.codec.decode(events[events.length - 1])
-  if (!event) return []
-  const version = events[events.length - 1]!.index
-  switch (event.type) {
-    case "PayerProfileUpdated":
-      const data = event.data
-      return [Upsert({ id: id, version, name: data.name, email: data.email })]
-    case "PayerDeleted":
-      return [Delete({ id: id, version })]
+  if (!state) return [Delete({ id, version })]
```
bartelink, line 18: hm actually deleting means replays would bring it back
but I guess making a tombstone to just to filter it out is a bit debatable too

packages/message-db/message-db/src/lib/Category.ts
```diff
@@ -199,7 +200,10 @@ export class MessageDbContext {
   }
 }
 
-type AccessStrategy<Event, State> =
+
+export type Project<State> = (conn: Client, streamId: Equinox.StreamId, state: State, version: bigint) => Promise<void>
```
bartelink, line 204: Project/projection is a very generic term
So is Index
But maybe it conveys "go off and make sure this is indexed" vs "maybe you want to project this somewhere somehow"
I guess I just really dislike Project
Maybe "ConsistentIndexing" or "InlineIndexing" ?

apps/example/src/read-models/PayerReadModel.ts
bartelink: ok, I should have remembered that!

apps/example/src/read-models/PayerReadModel.ts
nordfjord: but remember, there are no replays, this happens as events are appended in the same transaction 🎉

packages/message-db/message-db/src/lib/Category.ts
nordfjord: "InlineProjection" works, as discussed on discord

apps/example/src/read-models/PayerReadModel.ts
nordfjord: `projectionConfig`, `changes`?

apps/example/src/read-models/PayerReadModel.ts
bartelink: Yes, silly me!

apps/example/src/read-models/PayerReadModel.ts
bartelink: I think I prefer the former - changes works well as a name like you have in other places so better not to dilute that

packages/message-db/message-db/src/lib/Category.ts
nordfjord: But then... That kind of implies a single projection and this thing is a giant-ass footgun.
Maybe `FootgunProjection` 😜

bartelink: CouplingHook?

nordfjord: `UNSAFE_ProjectState<T>`

bartelink: `KimiProjection<T>`

PR #72, opened 2024-01-08
refactor: allow passing Events and Fold modules into config

nordfjord:
This makes for a slightly more approachable API as evidenced by the example project


bartelink:
Hm, if you're working from a cookbook and/or wrote the thing, having another named concept can make things better, and it does emphasize where each of the types come into play.

I'm not sure it works out for the best overall - sometimes things are just better travelling as two arguments and/or as tuples.

I guess the people will speak!

PR #75, opened 2024-01-10
feat: Codec.include

nordfjord:
the default json codec, while handy, will basically lie to you when it comes to decoding. As such we're providing a `Codec.include` function for scenarios where you don't necessarily want validation, but do want to ensure that you only decode known events.

```ts
type Event = { type: 'a' } | { type: 'b' } | { type: 'c' }
const codec = Codec.include(Codec.json<Event>(), [ 'a']) // type error, missing values
const codec = Codec.include(Codec.json<Event>(), [ 'a', 'b', 'c']) // all good

```

### DIFF
--- a/packages/core/src/lib/Codec.ts
+++ b/packages/core/src/lib/Codec.ts
@@ -39,6 +39,31 @@ export function json(mapMeta?: MapMeta<any, any>): ICodec<any, string, any> {
   }
 }
 
+type UnionToIntersection<U> =
+  (U extends any ? (k: U) => void : never) extends ((k: infer I) => void) ? I : never
+  
+type LastOf<T> =
+  UnionToIntersection<T extends any ? () => T : never> extends () => (infer R) ? R : never
+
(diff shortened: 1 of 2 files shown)


packages/core/src/lib/Codec.ts
```diff
@@ -39,6 +39,31 @@ export function json(mapMeta?: MapMeta<any, any>): ICodec<any, string, any> {
...
 
+type UnionToIntersection<U> =
+  (U extends any ? (k: U) => void : never) extends ((k: infer I) => void) ? I : never
+  
+type LastOf<T> =
+  UnionToIntersection<T extends any ? () => T : never> extends () => (infer R) ? R : never
+
+type Push<T extends any[], V> = [...T, V];
+
+type UnionToTuple<T, L = LastOf<T>, N = [T] extends [never] ? true : false> =
```
nordfjord, line 50: don't ask... type level programming is not my favourite thing

PR #76, opened 2024-01-13
feat: Support ttl for the cache

nordfjord:
Add support for ttl in the caching strategy. Allows us to keep the memory consumption down by purging stale items from the cache. Though, as it is an LRU, maybe people should just a lower `max`

### DIFF
--- a/packages/core/src/lib/Cache.ts
+++ b/packages/core/src/lib/Cache.ts
@@ -40,9 +40,15 @@ export interface ICache {
     skipReloadIfYoungerThanMs: number,
     supersedes: Supersedes,
     read: (tokenAndState?: TokenAndState<State>) => Promise<TokenAndState<State>>,
+    ttl?: number,
   ): Promise<TokenAndState<State>>
 
-  updateIfNewer<State>(key: string, supersedes: Supersedes, entry: CacheEntry<State>): void
+  updateIfNewer<State>(
+    key: string,
+    supersedes: Supersedes,
+    entry: CacheEntry<State>,
+    ttl?: number,
+  ): void
 }
 
 export class MemoryCache implements ICache {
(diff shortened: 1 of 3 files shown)


packages/core/src/lib/Cache.ts
```diff
@@ -97,10 +117,10 @@ export class MemoryCache implements ICache {
     return this.readWithExactlyOneFetch(key, () => read(current.value()))
```
bartelink, line 117: ttl missing?

nordfjord: Currently we don't have any ttl semantics on the cache, only LRU. Default size is 1M items which depending on the size of the state can be anywhere from a few megabytes when full up to a few gigabytes of memory usage. (a million kilobytes is a gigagbyte). So I think the question is whether to even have TTL support, or whether to ask people to configure sane size caps on their cache. JS doesn't have a way to inspect the byte size of an object either so there's no good way to determine the memory usage of the cache. There's no inherent harm in more memory usage either, if you're running on k8s you'll have configured a memory request and limit for your process, if you have at any point allocated `N` MB then that'll be reserved for your process going forward, so just because we free the memory doesn't actually mean another process can start using it. Lots of stuff to think about with this!

packages/propeller/src/StreamsSink.mts
```diff
@@ -7,9 +7,132 @@ import EventEmitter from "node:events"
...
 const tracer = trace.getTracer("@equinox-js/propeller")
 
+namespace StreamSpan {
+  const idx = (span: ITimelineEvent[]) => span[0]!.index
+  const ver = (span: ITimelineEvent[]) => span.at(-1)!.index
+  export function merge(min, spans: ITimelineEvent[][]) {}
+}
+
+namespace Buffer {
+  const WritePosUnknown = -2n
```
bartelink, line 17: this stuff is encoded like this in F# to stop the need for a bool for malformed pushing the memory requirement from 16bits or something similar - consider using flags like the original code did rather than having such vestigial nonsense!

packages/core/src/lib/Cache.ts
```diff
@@ -68,9 +68,8 @@ export class MemoryCache implements ICache {
   }
 
   private startPurgeTimer() {
-    setTimeout(() => {
+    setInterval(() => {
       this.cache.purgeStale()
-      this.startPurgeTimer()
     }, 5000).unref()
```
bartelink, line 73: maybe change 5000 to max (2000, min(5000, timeout/2)) ?
if people are feeling the memory pain this much then setting a TTL of 2s will worst case be 10s of memory consumption

nordfjord: FWIW I took the 5s number from microsofts memorycache that you're using in equinox

bartelink: I guess every traverse flushed CPU caches so you don't want it to be too frequently
but I also wouldnt read too much into the MS impl coz ultimately its dependent on how the GC interacts wth it
Habing said that, 5000 probably does make sense

packages/core/src/lib/Cache.ts
bartelink: 🤔

nordfjord:
Deciding against this for now, I'd rather tell people to constrain their cache to something reasonable than hack together a ttl cache.

PR #77, opened 2024-01-21
feat: Track stream positions to discard events and allow handlers to indicate how much of a stream they processed

nordfjord:
Inching closer to the propulsion streams sink as we speak. Added functionality to track stream write positions which allows us to discard events in at-least-once delivery scenarios that we know we've already processed.

Additionally we allow handlers to return results indicating how far they got. OverrideNextIndex is particularly useful in scenarios where you load the stream during processing. At that point you know that you've handled the stream up to index X and can tell propulsion "please don't call me with any events before X"


packages/propeller/src/Array.ts
```diff
@@ -0,0 +1,10 @@
+export function keepMap<T, V>(arr: T[], fn: (t: T) => V | undefined): V[] {
+  const result = new Array<V>(arr.length)
+  let idx = 0
+  for (let i = 0; i < arr.length; i++) {
+    const v = fn(arr[i])
+    if (v) result[idx++] = v
```
bartelink, line 6: could delay copying until first item is being dropped, or would you normally expect to be dropping something?

packages/propeller/src/Semaphore.ts
```diff
@@ -0,0 +1,16 @@
+export class Semaphore {
+  constructor(private readonly max: number) {}
+  private count = 0
+
+  tryTake(): boolean {
+    if (this.count < this.max) {
```
bartelink, line 6: if >= return false

packages/propeller/src/Sinks.ts
```diff
@@ -0,0 +1,32 @@
+import { ITimelineEvent } from "@equinox-js/core"
+
+export type StreamResult =
+  | { type: "NoneProcessed" }
+  | { type: "AllProcessed" }
+  | { type: "PartiallyProcessed"; count: number }
+  | { type: "OverrideNextIndex"; version: bigint }
```
bartelink, line 7: hm overridenext seems less confusing when you put it that way

packages/propeller/src/Array.ts
nordfjord: the most common case seems to be that it's called with one empty array and one array with some items. Mostly because overriding the write position is a `merge` with an empty queue

packages/core/src/lib/Decider.ts
```diff
@@ -101,6 +105,29 @@ export class Decider<Event, State> {
     )
   }
 
+  /**
+   * 1. Invoke the supplied `interpret` function with the present state to determine whether any write is to occur.
+   * 2. (if events yielded) Attempt to sync the yielded events to the stream.
+   *    (Restarts up to `maxAttempts` times with updated state per attempt, throwing `MaxResyncsExhaustedException` on failure of final attempt.)
+   * 3. Returns the new version of the stream
+   */
+  transactVersion(
```
bartelink, line 114: wondering whether I should rename my `TransactWithPostVersion` helpers - I guess this naming does align with the rest of the names....

packages/propeller/src/StreamsSink.mts
```diff
@@ -337,6 +337,8 @@ class ConcurrentDispatcher {
       // the span can be mutated during the computation, so we make a copy
       const events = span.slice()
 
+      // NOTE: this depends on the computation never throwing, otherwise we leak the semaphore
```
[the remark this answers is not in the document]
nordfjord: I think I should remove the comment. We don't do any error handling ATM and just bubble all errors up to the root causing a process crash. For now, that is the desired behaviour as we assume people to be running on environments where that'd cause a restart

packages/core/src/lib/Decider.ts
nordfjord: Given you're in a lang with function overloading and this is not, I'm not sure we should aim for function name parity. That said, `TransactWithPostVersion(decide)` is so long that us mere `fantomas` users might think twice before using it, so it's probably worth shortening it!

packages/propeller/src/StreamsSink.mts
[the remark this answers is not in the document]
bartelink: I think a note explaining the letting it crash would work just fine, no?

packages/core/src/lib/Decider.ts
bartelink: Yes function name parity is definitely not a goal
I'm just thinking the burden of overly long names should probably fall on people doing more fiddly and.or questionable things like checking input versions, whreas returning the post version is relatively normal and sane.

PR #80, opened 2024-02-17
feat: add a libsql category

nordfjord:


packages/sqlite/src/lib/LibSqlClient.ts
```diff
@@ -0,0 +1,145 @@
...
+      return parser(s)
+    }
+
+  export const row = (r: any): ITimelineEvent<Format> => {
+    const row = r as DbRow
+    const data = nullable(string)(row.data)
+    const meta = nullable(string)(row.meta)
+    const type = string(row.type)
+    const time = date(row.time)
+    const index = bigint(row.position)
```
bartelink, line 140: indx would align so pretty

packages/sqlite/src/lib/Read.ts
```diff
@@ -0,0 +1,93 @@
+import { ITimelineEvent } from "@equinox-js/core"
+import { Format, LibSqlReader } from "./LibSqlClient.js"
+import { trace } from "@opentelemetry/api"
+import { Tags } from "@equinox-js/core"
+
+type StreamEventsSlice = {
```
bartelink, line 6: there's maxbatches/batchSize naming - is slice meaningfully different? should it have been called that?
slice is kinda analogous to pages of events from across docs
maybe it's all fine but the historic term may be just that?

packages/sqlite/src/lib/Snapshot.ts
```diff
@@ -0,0 +1,27 @@
...
+import * as Token from "./Token.js"
+import { Format } from "./LibSqlClient.js"
+
+export const snapshotCategory = (original: string) => original + ":snapshot"
+
+export const streamName = (category: string, streamId: string) =>
+  `${snapshotCategory(category)}-${streamId}`
+export type Meta = { streamVersion: string }
+const streamVersion = (evt: ITimelineEvent<Format>) => {
+  const meta = JSON.parse(evt.meta ?? "null") as Meta | null
```
bartelink, line 11: does parsing the null literal achieve a type hint very cheaply?

packages/sqlite/src/lib/Snapshot.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const meta = (token: StreamToken): Meta => ({
+  streamVersion: String(Token.streamVersion(token)),
+})
+export async function decode<V>(
+  decode: (ev: ITimelineEvent<Format>) => Promise<V | null | undefined> | V | null | undefined,
+  events: ITimelineEvent<Format>[],
+): Promise<[StreamToken, V] | null> {
+  if (events.length > 0) {
+    const decoded = await decode(events[0])
```
bartelink, line 23: dont get the async - why?

packages/sqlite/test/libsql.test.ts
```diff
@@ -0,0 +1,502 @@
...
+    )
+    return Cart.Service.create(category)
+  }
+
+  export function createWithSnapshotStrategy(context: LibSqlContext) {
+    const access = AccessStrategy.AdjacentSnapshots<E, S>(
+      Cart.Fold.snapshotEventType,
+      Cart.Fold.snapshot,
+    )
+    const category = Category.create(context, Cart.Category, codec, fold, initial, noCache, access)
```
bartelink, line 49: CategoryName these days

packages/sqlite/src/lib/LibSqlClient.ts
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    try {
+      for (let i = 0; i < messages.length; ++i) {
+        const message = messages[i]
+        await trx.execute({
+          sql: `insert into messages(id, stream_name, category, type, data, metadata, position) 
+                values (?, ?, ?, ?, ?, ?, ?)`,
+          args: [
+            message.id || randomUUID(),
```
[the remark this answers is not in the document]
nordfjord: `import { randomUUID as genUUID } from "crypto"` doesn't feel quite right in the context of node

packages/sqlite/test/libsql.test.ts
nordfjord: This is just tests, in proper code it'd be `Cart.Stream.Category` which I think is specific enough given the namespace. I'm not quite sold on the `Name` postfix yet

packages/sqlite/src/lib/Snapshot.ts
nordfjord: Nope, in fact, this whole file should be removed because there's no snapshot mechanism yet

packages/sqlite/src/lib/Read.ts
nordfjord: oh good point, I think the naming comes from the F# version for SqlStreamStore and has just kind of followed along, haven't actually put time into thinking about the names as it's entirely clear to me what the type represents :sweat_smile:

packages/sqlite/src/lib/LibSqlClient.ts
nordfjord: you're serious aren't you?

packages/sqlite/src/lib/LibSqlClient.ts
```diff
@@ -0,0 +1,145 @@
+import { randomUUID } from "crypto"
+import type { Client, Transaction } from "@libsql/client"
+import { IEventData, ITimelineEvent } from "@equinox-js/core"
+
+type MdbWriteResult = { type: "Written"; position: bigint } | { type: "ConflictUnknown" }
```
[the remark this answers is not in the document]
nordfjord: :see_no_evil:

packages/sqlite/src/lib/Category.ts
```diff
@@ -0,0 +1,285 @@
...
+
+export type AccessStrategy<Event, State> = { type: "Unoptimized" } | { type: "LatestKnownEvent" }
+
+export namespace AccessStrategy {
+  export const Unoptimized = <E, S>(): AccessStrategy<E, S> => ({ type: "Unoptimized" })
+  export const LatestKnownEvent = <E, S>(): AccessStrategy<E, S> => ({ type: "LatestKnownEvent" })
+}
+
+class InternalCategory<Event, State, Context>
+  implements IReloadableCategory<Event, State, Context>
```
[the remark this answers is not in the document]
nordfjord: It's definitely a yuck, should just be part of the category interface itself and MemoryStore should implement it.

packages/sqlite/src/lib/Category.ts
```diff
@@ -0,0 +1,285 @@
...
+
+  async sync(
+    category: string,
+    streamName: string,
+    token: StreamToken,
+    encodedEvents: IEventData<Format>[],
+  ): Promise<GatewaySyncResult> {
+    const span = trace.getActiveSpan()
+    const streamVersion = Token.streamVersion(token)
+    const appendedTypes = new Set(encodedEvents.map((x) => x.type))
```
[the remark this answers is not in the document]
nordfjord: I think it's fine, you get the stream name on the span so it's easy to go from the trace to the DB to get the actual counts and order of types

packages/sqlite/src/lib/LibSqlClient.ts
bartelink: I've done that in the past and partially recovered my sanity, but have not forgotten the good times!

packages/sqlite/src/lib/Category.ts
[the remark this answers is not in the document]
bartelink: Sounds like you're spinning up the PR as we speak

packages/sqlite/src/lib/Category.ts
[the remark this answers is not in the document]
bartelink: yes, but it's very easy to have them in order of first appearance, which in general will be sufficient to correlate with code with 100% confidence without looking at anything
whreas unstable and inconsistent ordering is very confusing and can be misread as being different
Not letting this one go!

packages/sqlite/src/lib/Read.ts
bartelink: yeah, the batch term came from Eirik over ESDB
and is a central name in cosmos and ddb
I reckon batch is for the best to reduce term count by one

packages/sqlite/src/lib/Snapshot.ts
bartelink: mechanical sympathy of a kind eh. (like second guessing the hybrid unit in your rally car)

packages/sqlite/test/libsql.test.ts
bartelink: I removed Stream from the F# in full and life is quite good.
One driver is that Category clashes with lots of things in messy ways with shadowing (and Stream can potentially do that too)
Obv the Stream module and how it works in TS is good so Stream.Category vs CategoryName can be fine. Probably as long as Category is not used without the prefix, I guess the ambiguity is not to the same degree.

packages/sqlite/src/lib/Category.ts
```diff
@@ -0,0 +1,285 @@
...
+
+  async reload<Event, State>(
+    streamName: Equinox.StreamName,
+    token: StreamToken,
+    decode: Decode<Event>,
+    fold: (state: State, events: Event[]) => State,
+    state: State,
+  ): Promise<[StreamToken, State]> {
+    let streamVersion = Token.streamVersion(token)
+    const startPos = streamVersion + 1n // Reading a stream uses {inclusive} positions, but the streamVersion is `-1`-based
```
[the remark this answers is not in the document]
nordfjord: That +1 has to be somewhere

packages/sqlite/src/lib/Snapshot.ts
nordfjord: But now there's a snapshot mechanism!

packages/sqlite/src/lib/Category.ts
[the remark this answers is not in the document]
bartelink: -1 doesn't have to be generated or passed around.
Try to push it as far away as possible, so the span of code during which there is a mix of 0 based and -1 based is as small as possible
>= / > can often make it entirely redundant

packages/sqlite/src/lib/Category.ts
```diff
@@ -91,6 +92,26 @@ export class LibSqlContext {
...
+    streamName: string,
+    decode: Decode<Event>,
+    fold: (state: State, events: Event[]) => State,
+    initial: State,
+  ): Promise<[StreamToken, State]> {
+    const snapshot = await this.conn.read.readSnapshot(streamName)
+    if (snapshot == null) return [this.tokenEmpty, initial]
+    const decoded = decode(snapshot)
+    if (!decoded) return [this.tokenEmpty, initial]
+    const version = BigInt(snapshot.index)
```
bartelink, line 105: is this 0 or -1 based
same L110 L112
make them zero based so no confusion

packages/sqlite/src/lib/LibSqlClient.ts
```diff
@@ -59,6 +59,42 @@ export class LibSqlWriter {
...
+    event: IEventData<Format>,
+    expectedVersion?: bigint,
+  ): Promise<LibSqlWriteResult> {
+    const trx = await this.client.transaction("write")
+    try {
+      const current = await trx.execute({
+        sql: "select position from snapshots where stream_name = ?",
+        args: [streamName],
+      })
+      const currentVersion = current.rows.length === 0 ? -1n : BigInt(current.rows[0][0] as number)
```
bartelink, line 75: as noted earlier, a version on its own is not great for protecting in the face of multiple writers 
or maybe it is ;) just making sure you have some concurrency model in mind.

packages/sqlite/src/lib/Category.ts
```diff
@@ -191,13 +229,15 @@ export type AccessStrategy<Event, State> =
...
   | { type: "Snapshot"; isOrigin: (e: Event) => boolean; toSnapshot: (state: State) => Event }
+  | { type: "RollingState"; toSnapshot: (state: State) => Event }
 
 // prettier-ignore
 export namespace AccessStrategy {
   export const Unoptimized = (): AccessStrategy<any, any> => ({ type: "Unoptimized" })
   export const LatestKnownEvent = (): AccessStrategy<any, any> => ({ type: "LatestKnownEvent" })
   export const Snapshot = <E, S>(isOrigin: (e: E) => boolean, toSnapshot: (state: S) => E): AccessStrategy<E, S> => 
     ({ type: "Snapshot", isOrigin, toSnapshot })
+  export const RollingState = <E, S>(toSnapshot: (state: S) => E): AccessStrategy<E, S> => ({ type: "RollingState", toSnapshot })
```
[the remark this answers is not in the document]
nordfjord: the snapshot has an incrementing version (the version does not remain 0 in this implementation)

packages/sqlite/src/lib/LibSqlClient.ts
bartelink: I guess there's little point in ignoring the fundamentals of the backing store
But please tell me the version starts at 0 when there's no events!
A one liner at top of file might be good - want people scanning stores to understand that in general the version is the event count
and that rollingstate or custom means that there is an etag that further dictates whether you get to write just because you know the version
Also if you bake in the rolling state mode as the version in this way, you preclude ever having events - the Custom mode allows you to e.g. produce a checkpoint every 1000 updates or every hour etc

----

TL;DR if it was me, I think I'd have events be marked with a zero-based index like the rest of the world

and then in the snapshot table keep a tinyint etag thing that increments with wraparound

having said that, if not doing that means you can simplify, of course it's worth considering (but in my head I want to be able to easily move from defaulting to rollingsnapshot, and then deiding I want actual events for some things, without the version suddenly meaning something different entirely

Still 🤔 though

packages/libsql/src/lib/LibSqlClient.ts
```diff
@@ -2,20 +2,26 @@ import { randomUUID } from "crypto"
...
+  | { type: "Written"; position: bigint; snapshot_etag?: string }
+  | { type: "ConflictUnknown" }
 export type Format = string
 
 export class LibSqlWriter {
   constructor(private readonly client: Client) {}
 
+  /**
+   * Reads the current version of the stream
+   * version is 0 based, an empty stream has version 0
```
bartelink, line 15: so much unbridled joy and gratitude

packages/libsql/src/lib/LibSqlClient.ts
```diff
@@ -131,13 +137,14 @@ export class LibSqlReader {
     return result.rows.map(Parse.row)
   }
 
-  async readSnapshot(streamName: string) {
+  async readSnapshot(streamName: string): Promise<[string, ITimelineEvent] | undefined> {
     const result = await this.client.execute({
       sql: "select * from snapshots where stream_name = ?",
       args: [streamName],
     })
     if (result.rows.length === 0) return
```
bartelink, line 145: I wonder is there a linting rule you can set to require `return undefined` to be explicitly written to make people think?
Or is it just that in TS, thinking aint optional, ever!

packages/libsql/test/libsql.test.ts
```diff
@@ -225,73 +307,10 @@ describe("Round-trips against the store", () => {
...
-      s2()
-      const prepare2 = async () => {
-        s3()
-        await w4
-      }
-      await act(prepare2, service2, sku22, 22)
-    }
 
-    await Promise.all([t1(), t2()])
+    const { sku11, sku12, sku21, sku22, cartId } = await manufactureConflict(service1, service2)
```
bartelink, line 313: looking forward to the F# version already

packages/libsql/test/libsql.test.ts
```diff
@@ -529,13 +546,23 @@ describe("AccessStrategy.Snapshot", () => {
...
+        [Tags.loaded_count]: 0,
+        [Tags.cache_hit]: true,
+      },
     )
     memoryExporter.reset()
 
     // Add two more - the roundtrip should only incur a single read
     await CartHelpers.addAndThenRemoveItemsManyTimes(cartContext, cartId, skuId, service1, 1)
-    assertSpans({ name: "Transact", [Tags.loaded_count]: 0, [Tags.append_count]: 2, [Tags.cache_hit]: true })
+    assertSpans({
```
bartelink, line 560: maybe test code should have longer line len
maybe all code should have longer line len
maybe I should go write golang and learn tolerance

packages/sqlite/src/lib/Category.ts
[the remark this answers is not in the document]
nordfjord: lol, maybe next PR :sweat_smile:

packages/libsql/src/lib/LibSqlClient.ts
nordfjord: `return` is equivalent to `return undefined`, what's the problem?

packages/libsql/test/libsql.test.ts
nordfjord: uuurgh, yeah, fine... Unprettified the file and as expected it reads much better

bartelink: was only lamenting - it's really not the worst

packages/libsql/src/lib/LibSqlClient.ts
bartelink: Yes, was aware they are equivalent
Was just pondering that if it forced you to type `return undefined` that it would be more scannable
But it's not like sane scannability is in the top 10 concerns of JS, so why should it be for TS

packages/test-domain/src/domain/Cart.ts
```diff
@@ -35,7 +35,7 @@ export namespace Fold {
   export type State = { items: ItemInfo[] }
   export const toSnapshot = (s: State): Events.Compaction.State => ({ items: s.items })
   export const ofSnapshot = (s: Events.Compaction.State) => ({ items: s.items })
```
bartelink, line 37: ugh I need to unify the module snapshot and compaction terminology in the upstream
compaction was from Eirik's original terms in SaveForLater that he wrote as a bespoke thing, which I generalized to the loading backwards to an origin scheme

packages/libsql/test/libsql.test.ts
```diff
@@ -0,0 +1,561 @@
...
+      { name: "Query", [Tags.loaded_count]: 1, [Tags.cache_hit]: true },
+    )
+    expect(stale).toEqual(fresh)
+  })
+})
+
+test("Evolving access strategy from Unoptimised to Snapshot to RollingState", async () => {
+  const batchSize = 10
+  const context = createContext(client, batchSize)
+  const service1 = CartService.createWithoutOptimization(context)
```
bartelink, line 509: proper names 1/2/3 is unnecessary cryptic serviceU, sericeS, serviceR

packages/libsql/src/lib/LibSqlClient.ts
```diff
@@ -0,0 +1,203 @@
...
+
+type Trx = {
+  execute: Client['execute']
+}
+
+export class LibSqlWriter {
+  constructor(private readonly client: Client) {}
+
+  /**
+   * Reads the current version of the stream
```
bartelink, line 18: remove L18, making the important bit a one liner - maybe it goes in a header overview comment?
Looking forward to it being fixed in the messagedb impls :P

packages/libsql/src/lib/LibSqlClient.ts
```diff
@@ -0,0 +1,203 @@
...
+   * version is 0 based, an empty stream has version 0
+   */
+  async readStreamVersion(streamName: string, trx: Trx = this.client): Promise<bigint> {
+    const result = await trx.execute({
+      sql: "select max(position) from messages where stream_name = ?",
```
bartelink, line 23: would index be a better name now? aligns with i around the rest of the world, or does position have some specific meaning?

packages/libsql/src/lib/LibSqlClient.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  async writeMessages(
+    category: string,
+    streamName: string,
+    messages: IEventData<Format>[],
+    expectedVersion: bigint | null,
+    updateSnapshot?: (trx: Transaction) => Promise<void>,
+  ): Promise<LibSqlWriteResult> {
+    const trx = await this.client.transaction("write")
+    let position = await this.readStreamVersion(streamName, trx)
```
bartelink, line 40: version
```suggestion
    let version = await this.readStreamVersion(streamName, trx)
```

packages/libsql/src/lib/LibSqlClient.ts
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    try {
+      for (let i = 0; i < messages.length; ++i) {
+        const message = messages[i]
+        await trx.execute({
+          sql: `insert into messages(id, stream_name, category, type, data, metadata, position) 
+                values (?, ?, ?, ?, ?, ?, ?)`,
+          args: [
+            message.id || randomUUID(),
```
bartelink, line 53: is there not an earlier gate that can do this defaulting?
also why does schema not include a timestamp on the event

packages/libsql/src/lib/LibSqlClient.ts
```diff
@@ -0,0 +1,203 @@
...
+        INSERT INTO snapshots (stream_name, category, type, data, position, etag, id)
+        VALUES (?, ?, ?, ?, ?, ?, ?)
+        ON CONFLICT (stream_name) DO UPDATE
+        SET data = excluded.data, position = excluded.position, time = CURRENT_TIMESTAMP, type = excluded.type, id = excluded.id, etag = excluded.etag
+      `,
+        // prettier-ignore
+        args: [streamName, category, event.type, event.data ?? null, index, nextEtag, event.id ?? randomUUID()],
+      })
+      await trx.commit()
+      return { type: "Written", position: 0n, snapshot_etag: nextEtag }
```
bartelink, line 103: should position be index?
maybe nextEtag should be this or current?

packages/libsql/src/lib/LibSqlClient.ts
```diff
@@ -0,0 +1,203 @@
...
+    if (typeof s === "string") return s
+    throw new Error(`Expected string, got ${typeof s}`)
+  }
+  export const bigint = (s: unknown): bigint => {
+    switch (typeof s) {
+      case "string":
+      case "number":
+        return BigInt(s)
+      default:
+        throw new Error(`Expected string, got ${typeof s}`)
```
bartelink, line 176:
```suggestion
        throw new Error(`Expected string/number, got ${typeof s}`)
```

packages/libsql/src/lib/LibSqlClient.ts
```diff
@@ -0,0 +1,203 @@
...
+      return parser(s)
+    }
+
+  export const row = (r: any): ITimelineEvent<Format> => {
+    const row = r as DbRow
+    const data = nullable(string)(row.data)
+    const meta = nullable(string)(row.meta)
+    const type = string(row.type)
+    const time = date(row.time)
+    const index = bigint(row.position)
```
bartelink, line 198: but you used directly as a version in readVersion
I feel position is fine as a role name for arguments, but an unambiguous `index` is best for the schema. Though minIndex and maxIndex might be even better for that usage too (am thinking I want to go through all the stores and remove usage of the position name now)

packages/libsql/src/lib/Category.ts
```diff
@@ -0,0 +1,417 @@
...
+  ): Promise<[StreamToken, State]> {
+    const snapshot = await this.conn.read.readSnapshot(streamName)
+    // An end-user might add RollingState to a category after the fact, we should handle this gracefully
+    if (snapshot == null) return this.loadBatched(streamName, decode, fold, initial)
+    const [etag, event] = snapshot
+    const decoded = decode(event)
+    // An end-user might add RollingState to a category after the fact, we should handle this gracefully
+    if (!decoded) return this.loadBatched(streamName, decode, fold, initial)
+    const version = BigInt(event.index)
+    const state = fold(initial, [decoded])
```
bartelink, line 109: should you pass null instead? I know I dont do that consistently and it could be very frustrating to hit in prod, but if a random evolve from snapshot impl is assuming that the incoming state is initial (or something else), that's a recipe for a very long bug hunt and/or loss of trust

packages/libsql/src/lib/Category.ts
```diff
@@ -0,0 +1,417 @@
...
+    const span = trace.getActiveSpan()
+    span?.setAttributes({
+      [Tags.loaded_count]: 1,
+      [Tags.snapshot_version]: Number(version),
```
bartelink, line 113: hm back to prev discussion re version vs position vs index on snapshot
in cosmos, I have i which means "this is a snapshot taken when n was this value (n is the version and is the _n_ext index/i")
that would be valid as a version too (i.e. a snapshot with i=1 is indeed a snapshot at version 1)

I _think_ it all makes sense the way you have it, i.e. `index` on events is 0 based, and index on a snapshot is its logical index in the order (comes after the events, last event index+1)
that makes index on a snapshot be equivalent to a version.

OK so most of my other ramblings can prob be disregarded!

packages/libsql/src/lib/Category.ts
```diff
@@ -0,0 +1,417 @@
...
+    const snapshot = await this.conn.read.readSnapshot(streamName)
+    // An end-user might add snapshotting to a category after the fact, in which case there might not be a snapshot
+    // in all other cases the snapshot is guaranteed to exist if there are any events in the stream
+    if (snapshot == null) return this.loadBatched(streamName, decode, fold, initial)
+    const [_, event] = snapshot
+    const decoded = decode(event)
+    // The snapshot type may have changed, in which case we should ignore it and load a fresh state
+    if (!decoded || !isOrigin(decoded)) return this.loadBatched(streamName, decode, fold, initial)
+    const version = event.index
+    const state = fold(initial, [decoded])
```
bartelink, line 134: agian, consider passing null

packages/libsql/src/lib/Category.ts
```diff
@@ -0,0 +1,417 @@
...
+  ): Promise<[StreamToken, State]> {
+    const [version, events] = await Read.loadLastEvent(this.conn.read, streamName)
+    trace.getActiveSpan()?.setAttributes({
+      [Tags.loaded_count]: 1,
+      [Tags.read_version]: Number(version),
+    })
+    return [Token.create(version), fold(initial, keepMap(events, decode))]
+  }
+
+  async loadState<Event, State>(
```
bartelink, line 95: State doesnt convey much - either name it to convey somethign. about the stategy or add a comment to distinguish it from loadSnapshot (and maybe name that to convey that its loading for snapshot strategy more than just only literally loading a snapshot)

packages/libsql/src/lib/Category.ts
```diff
@@ -0,0 +1,417 @@
...
+  }
+
+  async reload<Event, State>(
+    streamName: Equinox.StreamName,
+    token: StreamToken,
+    decode: Decode<Event>,
+    fold: (state: State, events: Event[]) => State,
+    state: State,
+  ): Promise<[StreamToken, State]> {
+    let streamVersion = Token.version(token)
```
bartelink, line 150: stateVersion?

bartelink: side-note if the cached state is behind and the stream has snapshots, then doing a single read of that would be a better first step than loading potentially lots of events

I assume latestevent mode already does something itself to avoid this too (it cant in esdb/sss and I'm flaggign in case that logic is being carried forward suboptimally)

packages/libsql/src/lib/LibSqlClient.ts
bartelink: maybe position should be renamed to version and/or newVersion?

packages/libsql/src/lib/Category.ts
```diff
@@ -0,0 +1,417 @@
...
+    encodedEvents: IEventData<Format>[],
+    updateSnapshot?: (trx: Transaction) => Promise<void>,
+  ): Promise<GatewaySyncResult> {
+    const span = trace.getActiveSpan()
+    const streamVersion = Token.version(token)
+    const appendedTypes = new Set(encodedEvents.map((x) => x.type))
+    if (appendedTypes.size <= 10) {
+      span?.setAttribute(Tags.append_types, Array.from(appendedTypes))
+    }
+    const result = await this.conn.write.writeMessages(
```
bartelink, line 213: they are events man! (yes, I know why but still)

packages/libsql/src/lib/Category.ts
```diff
@@ -0,0 +1,417 @@
...
+  async sync(
+    category: string,
+    streamName: string,
+    token: StreamToken,
+    encodedEvents: IEventData<Format>[],
+    updateSnapshot?: (trx: Transaction) => Promise<void>,
+  ): Promise<GatewaySyncResult> {
+    const span = trace.getActiveSpan()
+    const streamVersion = Token.version(token)
+    const appendedTypes = new Set(encodedEvents.map((x) => x.type))
```
bartelink, line 209: see my earlier comments on main thread, avoidable bad news

packages/libsql/src/lib/Category.ts
```diff
@@ -0,0 +1,417 @@
...
+      t.state,
+    )
+    return { token, state }
+  }
+
+  private updateSnapshot(
+    category: string,
+    streamName: string,
+    ctx: Context,
+    newState: State,
```
bartelink, line 312: why the `new`?

packages/libsql/src/lib/Category.ts
```diff
@@ -0,0 +1,417 @@
...
+    }
+  }
+
+  async load(streamId: Equinox.StreamId, _maxStaleMs: number) {
+    const [token, state] = await this.loadAlgorithm(streamId)
+    return { token, state }
+  }
+
+  async reload(streamId: Equinox.StreamId, _requireLeader: boolean, t: TokenAndState<State>) {
+    if (this.access.type === "RollingState") {
```
bartelink, line 294: re earlier comment, for snapshot mode, cheapest reload is prob from that
maybe this is covered but naming and/or comments are not making it unambiguous in this fn

packages/libsql/src/lib/Category.ts
```diff
@@ -0,0 +1,417 @@
...
+    events: Event[],
+  ): Promise<SyncResult<State>> {
+    const span = trace.getActiveSpan()
+    const streamName = Equinox.StreamName.create(this.categoryName, streamId)
+    span?.setAttributes({
+      [Tags.category]: this.categoryName,
+      [Tags.stream_name]: streamName,
+    })
+    if (this.access.type === "RollingState") {
+      if (events.length === 0) return { type: "Written", data: { token, state } }
```
bartelink, line 347: how is it not a hard programmer error if we come in here with no events?

packages/libsql/src/lib/Category.ts
```diff
@@ -0,0 +1,417 @@
...
+import * as Read from "./Read.js"
+import { trace } from "@opentelemetry/api"
+import { Format, LibSqlReader, LibSqlWriter } from "./LibSqlClient.js"
+import { CachingCategory, ICachingStrategy, IReloadableCategory, Tags } from "@equinox-js/core"
+import { Client, Transaction } from "@libsql/client"
+import { randomUUID } from "crypto"
+
+const keepMap = Equinox.Internal.keepMap
+
+type GatewaySyncResult = { type: "Written"; token: StreamToken } | { type: "ConflictUnknown" }
```
bartelink, line 20: gateway is very old naming re esdb and trying to shoehorn in POESA naming
there is client, internalcategory and caterogy in play, maybe than naming can realign with that?
(PR accepted for doign that in upstream too)

packages/libsql/src/lib/Category.ts
nordfjord: I mean, it'll get short circuited in `transactAsync`, but I can't write the type as a `NonEmptyArray<T>` so better to just handle the case? Or would you want a throw here?

packages/libsql/src/lib/Category.ts
nordfjord: You're right, I'm only optimising RollingState here. Snapshot reload behaviour is the same as MessageDB

packages/libsql/src/lib/LibSqlClient.ts
nordfjord: the type of `IEventData` has `id` as an optional field, so this is the only place to do it really.

Yeah, timestamp is generated server side (avoids clock drift on many occasions)

packages/libsql/src/lib/Category.ts
nordfjord: I know F# allows you to do unsafe stuff like that, but TS does not. If I pass a `null` the compiler will yell at me because it is not a valid `State` value

packages/libsql/src/lib/Category.ts
nordfjord: They sure are, but I can also write commands into this table

packages/libsql/src/lib/Category.ts
bartelink: would go with a throw and a comment that it's for the compiler

packages/libsql/src/lib/LibSqlClient.ts
bartelink: could the ctor not fill it in?

packages/sqlite/src/lib/Category.ts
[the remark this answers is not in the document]
nordfjord: Changed to allowing up to 10, and if more we uniq them in order

packages/libsql/src/lib/LibSqlClient.ts
nordfjord: `arr[0]` = the first event... If it's index then the first event has index = 0 and the stream is at version = 1

Better to have something that conveys the first event is 1?

packages/libsql/src/lib/Category.ts
nordfjord: That was definitely forwarded onwards to mdb as well, will do a comprehensive optimisiation after merging this

packages/libsql/src/lib/Category.ts
nordfjord: Will do something more comprehensive later

packages/libsql/src/lib/LibSqlClient.ts
nordfjord: nvm, just changed it to be an actual index. Though some might argue index is 1 based and 0 based is offsets

packages/test-domain/src/domain/Cart.ts
nordfjord: Will do something about it after merge

PR #82, opened 2024-02-26
Optimise reloads

nordfjord:
- perf(message-db): optimise reloads
- perf(libsql): optimise reloads

### DIFF
--- a/packages/libsql/src/lib/Category.ts
+++ b/packages/libsql/src/lib/Category.ts
@@ -53,15 +53,15 @@ export class LibSqlContext {
     decode: Decode<Event>,
     fold: (state: State, events: Event[]) => State,
     initial: State,
+    version: bigint,
   ): Promise<[StreamToken, State]> {
     let state = initial
-    let version = 0n
     for await (const [lastVersion, events] of Read.loadForwardsFrom(
       this.conn.read,
       this.batchSize,
       this.maxBatches,
       streamName,
-      0n,
+      version,
     )) {
       state = fold(state, keepMap(events, decode))
       version = lastVersion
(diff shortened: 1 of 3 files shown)


packages/libsql/src/lib/Category.ts
```diff
@@ -252,18 +253,25 @@ class InternalCategory<Event, State, Context>
   }
 
   async reload(streamId: Equinox.StreamId, _requireLeader: boolean, t: TokenAndState<State>) {
-    if (this.access.type === "RollingState") {
-      return this.load(streamId, 0)
+    switch (this.access.type) {
+      // These strategies all have it in common that their state 
+      // can be reconstituted in a single round trip to the database
+      // as such is it as cheap to load them as it is to reload them
```
nordfjord, line 259: possible caveat on snapshot, where the reload could be sending more data over the wire in cases where the cached value is the same as what's in the cache

bartelink, line 259:
```suggestion
      // as such, is it as cheap to load them as it is to reload them
```
But maybe add even more explicit stuff?
// for this reason, unlike with some other stores, there's no particular win to be had in terms of
// using a more effficient index based on some cached token or etag, reducing roundtrips etc.

packages/libsql/src/lib/Category.ts
```diff
@@ -252,18 +253,25 @@ class InternalCategory<Event, State, Context>
...
+      case "LatestKnownEvent":
+        return this.load(streamId, 0)
+      case "Unoptimized":
+        const streamName = Equinox.StreamName.create(this.categoryName, streamId)
+        const [token, state] = await this.context.loadBatched(
+          streamName,
+          this.codec.decode,
+          this.fold,
+          t.state,
+          Token.version(t.token),
```
bartelink, line 271: maybe a helper for getting `[token, state]` from `t` would make things clearer (if there's only one or two never mind; this is obviously not exactly cryptic)

packages/message-db/message-db/src/lib/Category.ts
```diff
@@ -194,6 +194,8 @@ class InternalCategory<Event, State, Context>
   private async loadAlgorithm(
     streamId: Equinox.StreamId,
     requireLeader: boolean,
+    initial: State,
+    version = 0n,
```
bartelink, line 198: not sure about defalt arg as you didnt do it for loadbatched in the other impl

nordfjord: fair, went through and reduced the line count through "uglification" as well

packages/message-db/message-db/src/lib/Category.ts
```diff
@@ -206,87 +197,43 @@ class InternalCategory<Event, State, Context>
...
-        ;[token, state] = await this.context.loadBatched(
-          streamName,
-          requireLeader,
-          this.codec.decode,
-          this.fold,
-          t.state,
-          Token.version(t.token),
-        )
+        // prettier-ignore
+        result = await this.context.loadBatched(streamName, requireLeader, this.codec.decode, this.fold, t.state, Token.version(t.token))
```
bartelink, line 233: could the last 2 args become a TokenAndState?

PR #86, opened 2024-12-11
fix docs: events in past tense

mpranjic:


docs/docs/dynamo-store/access-strategies.md
```diff
@@ -31,8 +31,8 @@ In Equinox the Snapshot is a member of the Event DU
 
 ```ts
 type Event =
-  | { type: "Increment"; data: { amount: number } }
-  | { type: "Snapshot"; data: { current: number } }
+  | { type: "Incremented"; data: { amount: number } }
```
nordfjord, line 34: update in fold below as well

mpranjic: I think I got all of them now

PR #94, opened 2026-04-07
fix: use consumerGroupMember in groupName

nordfjord:
resolves #93

appends `-{groupMember}` to the groupName before passing it on.

This will require a major version update since this bug has existed for well over 2 years

### DIFF
--- a/packages/message-db/message-db-consumer/src/lib/MessageDbSource.mts
+++ b/packages/message-db/message-db-consumer/src/lib/MessageDbSource.mts
@@ -95,8 +95,12 @@ export class MessageDbSource {
   static create(options: Options & { pool: Pool }) {
     const client = new MessageDbCategoryReader(options.pool)
     const sink = options.sink
+    let groupName = options.groupName
+    if (options.consumerGroupMember != null) {
+      groupName = `${groupName}-${options.consumerGroupMember}`
+    }
     sink.addTracingAttrs({
-      "eqx.consumer_group": options.groupName,
+      "eqx.consumer_group": groupName,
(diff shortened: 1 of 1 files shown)


packages/message-db/message-db-consumer/src/lib/MessageDbSource.mts
```diff
@@ -95,8 +95,12 @@ export class MessageDbSource {
   static create(options: Options & { pool: Pool }) {
     const client = new MessageDbCategoryReader(options.pool)
     const sink = options.sink
+    let groupName = options.groupName
+    if (options.consumerGroupMember != null) {
+      groupName = `${groupName}-${options.consumerGroupMember}`
```
lostplan, line 100: I'd be tempted to include `consumerGroupSize` in the key to force users to consider checkpoints when resizing the group

nordfjord: Looking at what eventide does when [consumer groups](https://docs.eventide-project.org/user-guide/consumers.html#consumer-groups) are used. They say

> [!CAUTION]
>
> Consumers operated in consumer groups must be used in conjunction with the identifier attribute, or else the individual consumers in a consumer group will overwrite each other's [position records](https://docs.eventide-project.org/user-guide/consumers.html#position-store).

I'm tempted to take that same approach in equinox, place the burden entirely in user land. I think modifying the groupName inside equinox probably makes for surprises for the end-user when e.g. they look at their spans and see something other than they put in in the consumer_group attribute. Or if they define the groupName as a constant somewhere in their code and it doesn't match, etc.

PR #91, opened 2026-03-31
docs/update

nordfjord:
- **docs: upgrade marketing**
- **review fixes**

PR #96, opened 2026-04-13
apply bartelinks suggestions from #91

nordfjord:


docs/docs/faq.md
```diff
@@ -22,13 +22,13 @@ and replay semantics, this style tends to age well.
 
 ## Who is it not for?
 
-EquinoxJS is probably not a good fit if you want the library to hide the shape
-of event sourcing from you.
+EquinoxJS is probably not a good fit if your aim is to hide the shape of event
+sourcing from you rather than learn and apply it directly.
```
nordfjord:
```suggestion
EquinoxJS is probably not a good fit if your aim is to hide the shape of event
sourcing rather than learn and apply it directly.
```

docs/docs/dynamo-store/architecture.md
```diff
@@ -58,6 +58,10 @@ The key point is that DynamoStore is not a generic adapter layer. It is an event
 store design tuned specifically for DynamoDB so EquinoxJS can preserve explicit
 stream semantics in a serverless environment.
 
+That tuning stays at the store and category boundary. It does not leak into
+domain code or force you to turn services upside down if you later choose a
+different store.
```
bartelink, line 63: or you opt to migrate one category to a different store, store instance etc - i.e. convey that each one is as independent a spec as it should be vs having a single global config for an app that needs to define all config for all categories in a single thing

docs/docs/message-db/index.md
```diff
@@ -17,14 +17,14 @@ models, reactions and access strategies can still be tuned explicitly.
...
-thing.
+thing, or asking your application team to design and maintain the gap.
 
 Use MessageDB when you want:
 
 - PostgreSQL as the operational substrate
 - category-based consumers and checkpoints
 - store-supported inline projections when needed
-- a mature stream-store model rather than a generic event table
+- a mature stream-store model rather than owning a generic events table yourself
```
bartelink, line 27: owning here sounds like an asset more than conveying that you have a responsivility to architect, troubleshoot and tune in perpetuity

docs/docs/message-db/index.md
```diff
@@ -17,14 +17,14 @@ models, reactions and access strategies can still be tuned explicitly.
 If your default instinct is "we already run Postgres", this is the page to look
 at first. MessageDB gives EquinoxJS a straightforward PostgreSQL deployment
 story without pretending a generic events table and a stream store are the same
-thing.
+thing, or asking your application team to design and maintain the gap.
```
bartelink, line 20: "the gap" doesnt speak to me. Maybe something should convey that event processing infra and sql designs need tests and are non trivial. a working maintainable event storage system is more than a CREATE statement

docs/docs/faq.md
```diff
@@ -52,7 +52,8 @@ No. That is deliberate.
 
 EquinoxJS gives you the domain and store primitives needed to build a service,
 projector or process manager without prescribing transport, hosting or handler
-shape.
+shape. It is also not intended to accrete framework-style features every time a
```
bartelink, line 55: accrete us very good thesuarus use, but I guess I was alluding to the fact that you're not waiting for the framework to sprout a new feature the minute you have a use case that doesnt fit the predefined facilities

docs/src/components/HomepageFeatures/index.js
```diff
@@ -53,7 +53,8 @@ const FeatureList = [
     description: (
       <>
         Reactions and projection helpers cover the common read-model and process-manager patterns
-        without locking you into one shape.
+        without locking you into one shape, so the same model can handle both the trivial cases and
```
bartelink, line 56:
```suggestion
        without locking you into one shape, so the same model can be applied to both the trivial cases and
```

